PR #140, opened 2015-03-12
Adds scalar support to XML deserialization.

mattvryan:
This change reads the input string from the Reader in the IOContext supplied to the FromXmlParser to try to determine if the input XML represents a scalar value.  If so, when asking for the nextToken(), this change skips the tokens that don't pertain to scalars to get to the scalar value contained inside the XML.  This shouldn't impact other XML strings.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/deser/FromXmlParser.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/deser/FromXmlParser.java
@@ -1,6 +1,8 @@
 package com.fasterxml.jackson.dataformat.xml.deser;
 
+import java.io.BufferedReader;
 import java.io.IOException;
+import java.io.StringReader;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


cowtowncoder:
I don't think the code will work; while during testing `StringReader` is often used, this is not guaranteed. More commonly for real use, `InputStream` would be used.
Other possibly problem is use of hard-coded element names, although I guess that might be ok.
Fundamentally working around xml parser with straight string matching also seems like the wrong way, considering that one could have comments, white space and/or processing instructions preceding root value.

mattvryan:
Well, I understand.  It was a long shot anyway.

Based on your concerns, I think we'll have to give this one up.  I've been up and down through the entire call stack, and there's never really a place where all of the information needed to perform this function is available at the same time.  This seemed to be the best chance.

The last two issues you mentioned could possibly be handled, although they'd probably be ugly and inefficient.  But if we can't rely on the underlying of the reader supporting mark() and reset() etc. then this approach is a non-starter - and looking through the call stack I'm not seeing another way that will work.

So thanks anyway.  I'll close everything out.

cowtowncoder:
Sorry to hear that. And I appreciate your attempts here -- I agree, it is difficult to pull off.

I will still keep this in back of my mind, as I have noticed that occasionally opportunities open up.
Perhaps in your case you can handle it at higher level, using rewindable input source in the meantime.

PR #157, opened 2015-08-07
Update readme, fix example code error in incremental writes section.

zacyang:
Hi , I tried your lib this afternoon, when I following the example code there's an error, which I think maybe misleading for the newbies of this lib.
Therefor create the PR.


cowtowncoder:
Thank you, much appreciated!

Issue #172, opened 2015-11-21
XML INDENT_OUTPUT property fails to add newline/indent initial elements


rpatrick00:
I am using Jackson 2.6.3 with JAXB annotations using the following code:

<pre>
    XmlMapper mapper = new XmlMapper();

    mapper.registerModule(new JaxbAnnotationModule());
    mapper.configure(SerializationFeature.INDENT_OUTPUT, true);
    mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
    mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
    mapper.configure(SerializationFeature.WRITE_EMPTY_JSON_ARRAYS, false);
    mapper.setSerializationInclusion(JsonInclude.Include.NON_EMPTY);
    mapper.configure(ToXmlGenerator.Feature.WRITE_XML_1_1, true);
    mapper.writeValue(fileOutputStream, myCompany);
</pre>

The file I get looks like the following.  How do I get Jackson to add the newlines and missing indentation?

<pre>
&lt;?xml version='1.1' encoding='UTF-8'?&gt;&lt;company&gt;&lt;employees&gt;
    &lt;employee id="emp-1"&gt;
      &lt;name&gt;Robert Patrick&lt;/name&gt;
      &lt;type&gt;FULLTIME&lt;/type&gt;
    &lt;/employee&gt;
    &lt;employee id="emp-2"&gt;
      &lt;name&gt;Michael Smith&lt;/name&gt;
      &lt;type&gt;CONTRACTOR&lt;/type&gt;
    &lt;/employee&gt;
  &lt;/employees&gt;
&lt;/company&gt;
</pre>

The Java classes look like this.

Company.java:
<pre>
@XmlRootElement(name = "company")
@XmlAccessorType(XmlAccessType.FIELD)
public class Company {

  @XmlElementWrapper(name = "employees")
  @XmlElement(name = "employee")
  private List<Employee> employees;

  public Company() {
    employees = new ArrayList<Employee>();
  }
...
}
</pre>

Employee.java:
<pre>
@XmlRootElement(name = "employee")
@XmlAccessorType(XmlAccessType.FIELD)
public class Employee {
  @XmlAttribute
  @XmlID
  private String id;

  @XmlElement
  private String name;

  @XmlElement
  private EmployeeType type;
  ...
}
</pre>

EmployeeType:
<pre>
@XmlEnum
[...]

PR #173, opened 2015-11-23
Fixing issue #172

rpatrick00:
This fix seems to work for me but I was unable to get unit tests to work in my environment before making any code changes so I have no feedback on whether it broke anything else.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java
@@ -30,6 +30,8 @@ import com.fasterxml.jackson.dataformat.xml.util.StaxUtil;
 public final class ToXmlGenerator
     extends GeneratorBase
 {
+    private final static String NEWLINE = System.getProperty("line.separator", "\n");
+
     /**
      * If we support optional definition of element names, this is the element
      * name to use...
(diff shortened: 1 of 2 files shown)


rpatrick00:
These are the same errors I was seeing before any changes...I don;t think my change could have such sweeping impact such as:

java.lang.NoClassDefFoundError: Could not initialize class com.fasterxml.jackson.databind.SerializationConfig

cowtowncoder:
This could be a version discrepancy in dependencies. I'll have a look.

src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java
```diff
@@ -195,8 +197,14 @@ public void initGenerator()  throws IOException
         try {
             if ((_xmlFeatures & Feature.WRITE_XML_1_1.getMask()) != 0) {
                 _xmlWriter.writeStartDocument("UTF-8", "1.1");
+                if (_xmlPrettyPrinter != null) {
+                  _xmlWriter.writeRaw(NEWLINE);
```
cowtowncoder: This may work; my main concern is that it is not supported by Stax implementation included by JDK.
As this would throw an exception (for non-implemented), I'll have to think about this.
As per my notes on bug, there is no reliably way to inject whitespace outside of root element, unless raw writes (which by-pass underlying XML encoder) are supported.

rpatrick00:
What about adding a newline at the end of the XML header then?

cowtowncoder:
@rpatrick00 XML declaration is written via method that takes optional arguments for encoding, version, but it does not take literal String. So there is no way to add linefeed that way.

I think your fix for the other indentation is correct, regardless; I'll start with that.

cowtowncoder:
Thank you for both reporting this, and showing how to fix -- I did slightly different take, but the idea is same. Only addition is that there's also trailing linefeed after content, in case of indentation; I think that makes sense.

Issue #162, opened 2015-10-02
XML Empty tag to Empty string in the object during xml deserialization


thefalconfeat:
I am using jackson-dataformat-xml:2.6.2

If an empty tag is available in the xml and the respective field in the object is a string, it would be nice to have jackson populate the field with empty string rather than with null value.

Ex:

``` java
@XmlAccessorType(XmlAccessType.FIELD)
@XmlType(
        name = "",
        propOrder = {"first", "last"}
)
@XmlRootElement(
        name = "name"
)
public class Name {
private String first;
private String last;
}
```

xml:

``` xml
<name>
<first>Ryan</first>
<last></last>
</name>
```

mapper creation:

```
new XmlMapper(XMLInputFactory2.newFactory(), XMLOutputFactory2.newFactory());
```

I would expect 

```
Name name = mapper.readValue(xml, Name.class)
```

name.getLast().equals("") to return true but name.getLast() ends up being null.

Any help would be appreciated.

cowtowncoder:
Yes I can reproduce this. It is a tricky problem to solve but I added a failing unit test so that hopefully it may be resolved in future.

In the meantime a work-around would be to add explicit `setLast()` that can convert `null` into empty String (and perhaps also pre-initialize `last` as empty string in constructor.

cowtowncoder:
I have some ideas as to how this could be implemented; but actual fix will need to wait until 2.8 unfortunately, as some API changes will be needed.

RobinAllen:
I'm also facing what I think is the same issue.   
In my case (simplified here to just the basic problem), I'm reading xml which defines a 'Thing's state as A or B:

State A : `"<Thing><A /></Thing>"`
State B : `"<Thing><B /></Thing>"`

Both A and B are defined in XSD as being 'xsd:string' with max length 0 characters. 'Thing' is a complextype, defined as being <xsd:choice> between element A or element B. They are _intended_ to have no content; it is purely the _presence_ or _absence_ of the empty element which declares which state we are in. Sadly I don't have control over the interface to be able to change the logic.  I have to live with this XML interface.

In Java, the POJO generated from XSD has String fields 'a' and 'b'.

If I serialize from POJO to XML I have no problem, but the trouble comes when I try to deserialize again.

If I deserialize `"<Thing><A /></Thing>"`, then I want to end up with [ a="", b=null ]
If I deserialize `"<Thing><B /></Thing>"`, then I want to end up with [ a=null, b="" ]
[...]

PR #218, opened 2016-12-22
Fix for #162 (XML Empty tag to Empty string)

oistein:
Fix for issue #162.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/deser/XmlTokenStream.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/deser/XmlTokenStream.java
@@ -371,21 +371,23 @@ public class XmlTokenStream
     
     private final String _collectUntilTag() throws XMLStreamException
     {
-        String text = null;
+        if (_xmlReader.isEmptyElement()) {
+            _xmlReader.next();
+            return null;
+        }
+
+        StringBuilder text = new StringBuilder();
+
         while (true) {
             switch (_xmlReader.next()) {
             case XMLStreamConstants.START_ELEMENT:
             case XMLStreamConstants.END_ELEMENT:
             case XMLStreamConstants.END_DOCUMENT:
-                return text;
-                // note: SPACE is ignorable (and seldom seen), not to be included
+                return text.toString();
+            // note: SPACE is ignorable (and seldom seen), not to be included
             case XMLStreamConstants.CHARACTERS:
             case XMLStreamConstants.CDATA:
-                if (text == null) {
-                    text = _xmlReader.getText();
-                } else { // can be optimized in future, if need be:
-                    text += _xmlReader.getText();
-                }
+                text.append(_xmlReader.getText());
                 break;
             default:
                 // any other type (proc instr, comment etc) is just ignored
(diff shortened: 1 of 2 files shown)


oistein:
Any updates on this?

Mo-la-machette:
Hello everyone ! Is this pull request going to be merged soon? thank you

src/main/java/com/fasterxml/jackson/dataformat/xml/deser/XmlTokenStream.java
```diff
@@ -371,21 +371,23 @@ private final int _next() throws XMLStreamException
     
     private final String _collectUntilTag() throws XMLStreamException
     {
-        String text = null;
+        if (_xmlReader.isEmptyElement()) {
+            _xmlReader.next();
+            return null;
+        }
+
+        StringBuilder text = new StringBuilder();
```
cowtowncoder: Hmmh. Why? This was an optimization for the common case of just single segment, in which case construction of intermediate `StringBuilder` was avoided. It's probably not a big deal either way, but usually I prefer not fixing things that ain't broke.

cowtowncoder: Oh. Actually, I see. So, this was to get empty String over `null`.
I think I'll modify this bit differently, after first merging,.

cowtowncoder:
Ok, I am really torn here. The challenge here is this: fix changes low-level behavior, regardless of what target type might be. This means that it could change handling of empty/null cases for any number of other scalar types; here we'd prefer only changing `String` case (which I think should ALWAYS get to empty String). In addition, distinction between empty XML element and start/end should not have any significance, from XML perspective: logically the two are equivalent.

Still.. I think I'm ok for trying this out for 2.9.0, and see if it works better than before. I may be sorry (I've had a few similar-looking cases that burnt :) )... but we'll see.

cowtowncoder:
Ended up implementing #245 and #246 that should help here: so, by default empty element (but NOT start+end with no content) will be exposed as `JsonToken.VALUE_NULL`. But this may changed via `FromXmlParser.Feature.EMPTY_ELEMENT_AS_NULL` to instead return `JsonToken.VALUE_STRING` (with text of "").
Enabling this feature is likely most natural way to deal with XML content, and doing that will also help with things like "empty" Objects (instead of `null` reference).

mliberty:
It would be nice to also have a way to make start+end with no content deserialize to `null`, especially since that was previous behavior.

cowtowncoder:
@mliberty these combinations are bit of a mess. But I didn't think start+end did deserialize to `null` previously?

I am now bit torn as to whether to also add matching feature for START+END; but that should go in 3.0.

Also, there seem to be (at least?) 3 distinct possibilities for both:

1. Expose as "native" `null`
2. Expose as empty String value ""
3. Exposes as logical `START_OBJECT`/`END_OBJECT`

all of which make sense for some use cases.

Although due to wide variety of coercions available, perhaps (3) is not needed: it is bit trickier to support (need to emit 2 tokens; will confuse functionality at databind level).

cowtowncoder:
Unfortunately with 2.x there is no way to make all cases work well: it would be easy enough to induce `null` for start+end, but that would probably break some code that expects empty String.
With "null coercion" many things are possible, but in patch release such functional changes should not be added.

3.0.0 will now implement this, however, as suggested.

PR #220, opened 2017-01-19
Avoid root-level wrapping for Map-like types, not just Maps

### LINKED ISSUE #5: Map serialization adds extra element
Current Map serializer produces structure like:

<bean>
  <mapProperty>
    <mapProperty>
     <key1>value1</key1>
     <key2>value2</key2>
    </mapProperty>
  </mapProperty>
</bean>

where there is one too many 'mapProperty' elements.

ruslansennov:
Custom map-like types is still wrapped (see [this](https://github.com/ruslansennov/javaslang-jackson/blob/fa5d70f37006eb94f705fe6f12aaccb63cfe39f4/src/test/java/javaslang/jackson/datatype/map/MapTest.java#L126) failed test)

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/TypeUtil.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/TypeUtil.java
@@ -21,7 +21,7 @@ public class TypeUtil
                 return false;
             }
             // issue#5: also, should not add wrapping for Maps
-            if (Map.class.isAssignableFrom(cls)) {
+            if (type.isMapLikeType()) {
                 return false;
             }
             return true;


cowtowncoder:
Quick question: what do you mean by "fixed #5"? It looks like this might fix issues with Scala `Map`s, but I want to make sure I understand the context, since I don't see new unit tests.

ruslansennov:
> what do you mean by "fixed #5"?

My fail test shows these results of serialization to xml:

```xml
Java:
<xmlSerialize xmlns=""><transitType><key1>1</key1><key2>2</key2></transitType></xmlSerialize>
Javaslang:
<xmlSerialize xmlns=""><transitType><transitType><key1>1</key1><key2>2</key2></transitType></transitType></xmlSerialize>
```

So, I believe here we have the same problem like in #5

cowtowncoder:
Ok thanks! I did manual merge via 2.8 branch to get in 2.8.7; same fix.

PR #223, opened 2017-02-14
Fix #222

### LINKED ISSUE #222: DefaultXmlPrettyPrinter indentation configuration not preserved
I want to prettyprint XML with a `\n` character, disregarding the system dependent line feed/carriage return settings.

I thought that the following code would be sufficient:

```java
XmlMapper xmlMapper = new XmlMapper();
DefaultXmlPrettyPrinter pp = new DefaultXmlPrettyPrinter();
pp.indentObjectsWith(new Newline2SpacesIndenter());
ObjectWriter xmlWriter = xmlMapper.writer(pp);
```

Where `Newline2SpacesIndenter()` is my custom indenter.

However, attempting to use the resulting ObjectWriter doesn't result in the expected behaviour - the customized indenter is lost, as `DefaultXmlPrettyPrinter.createInstance()` doesn't preserve configuration.

Is this a bug, or am I attempting to configure things in the wrong place?

jtuc:
Ensure DefaultXmlPrettyPrinter configuration is preserved across instances

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
@@ -125,7 +125,11 @@ public class DefaultXmlPrettyPrinter
     
     @Override
     public DefaultXmlPrettyPrinter createInstance() {
-        return new DefaultXmlPrettyPrinter();
+        DefaultXmlPrettyPrinter instance = new DefaultXmlPrettyPrinter();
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Thank you! I think I'll backport this in earlier version(s), so will do manual merge. 
I agree, settings should be retained; I think this is a regression of an earlier refactoring.
I'll see if I can create a unit test to verify it.

cowtowncoder:
@jtuc Thank you again for reporting the problem, suggesting fix -- I backported in 2.7(.10), 2.8(.7).

jtuc:
Thanks very much for the quick response!

PR #232, opened 2017-05-09
Adding writeRawValue method support to ToXmlGenerator.

Falland:
Currently only writeRaw methods are supported.
Default implementation for writeRawValue from GeneratorBase redirects call to writeRaw.
This lead to element name to be ignored which in order leads to invalid XML.

This change will allow to write 'writeRawValue' code in the same fashion as Json one and generating valid XML

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java
@@ -655,6 +655,82 @@ public final class ToXmlGenerator
      */
 
     @Override
+    public void writeRawValue(String text) throws IOException {
+        // [dataformat-xml#39]
+        if (_stax2Emulation) {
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Sounds good. Would it be possible to add unit tests to show that this works?
I'd be happy to merge this.

One other thing: before merge, we need Contributor Licenses Agreement (CLA):

(unless we have asked and gotten one before -- only ever need to  be done once)

Usual way is to print, fill & sign, scan, email to `info` at fasterxml dot com. Once this is done contribution may be merged.

Looking forward to merging this patch, thank you for submitting it!

Falland:
Hi,

Thanks for quick reply.
I've added another commit with the tests.
Ad I've also filled and sent the form.

Looking forward to see this change merged in repo.

Kind regards,

Yury

cowtowncoder:
Excellent, thank you! Will merge.

Issue #233, opened 2017-05-16
XmlMapper.copy() doesn't properly copy internal configurations


gtrog:
Affected version: all versions (up to jackson-dataformat-xml-2.9.0.pr3)

Currently, if you create an instance of XmlMapper and register a module with it, any copies of that mapper will not contain the registered module.  This behavior is different from the behavior of `ObjectMapper`, which retains all other configurations, such as modules, serialization/deserialization configs, etc.

```
final XmlMapper mapper = new XmlMapper();
mapper.registerModule(new JaxbAnnotationModule());

final XmlMapper mapperWithoutModule = mapper.copy();
```

I think this could be easily resolved by following the same copy-constructor pattern that `ObjectMapper` uses to copy itself, and then just invoking super's constructor in `XmlMapper`'s copy constructor.

I'll have a PR up shortly.

cowtowncoder:
Thank you for reporting this: yes, copy() should retain module settings.

I added one important comment; I think serializer provider does need to be copied. Otherwise LGTM.

cowtowncoder:
Thank you! I also backported this in 2.8, to be included in 2.8.9.

PR #234, opened 2017-05-16
Issue#233 XmlMapper.copy() doesn't properly copy internal configurations

gtrog:
Resolves issue #233

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/XmlMapper.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/XmlMapper.java
@@ -1,22 +1,20 @@
 package com.fasterxml.jackson.dataformat.xml;
 
-import com.fasterxml.jackson.core.PrettyPrinter;
-import com.fasterxml.jackson.core.Version;
-import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.AnnotationIntrospector;
-import com.fasterxml.jackson.databind.JavaType;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.dataformat.xml.deser.FromXmlParser;
-import com.fasterxml.jackson.dataformat.xml.ser.ToXmlGenerator;
-import com.fasterxml.jackson.dataformat.xml.ser.XmlSerializerProvider;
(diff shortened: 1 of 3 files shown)


src/main/java/com/fasterxml/jackson/dataformat/xml/ser/XmlSerializerProvider.java
```diff
@@ -56,6 +55,11 @@ public XmlSerializerProvider(XmlSerializerProvider src,
      */
 
     @Override
+    public DefaultSerializerProvider copy() {
+        return this;
```
gtrog: I can add a comment if needed, but I'm doing this because `XmlSerializerProvider` is effectively immutable in that it's only member variable is also immutable, and so I can avoid having to try to figure out whether I should construct it with just an instance of `XmlRootNameLookup` or with an instance of itself and `SerializationConfig` and `SerializerFactory`

src/main/java/com/fasterxml/jackson/dataformat/xml/XmlMapper.java
```diff
@@ -1,21 +1,23 @@
...
 
-import java.io.IOException;
-
-import javax.xml.stream.XMLInputFactory;
-import javax.xml.stream.XMLOutputFactory;
-import javax.xml.stream.XMLStreamReader;
-import javax.xml.stream.XMLStreamWriter;
-
-import com.fasterxml.jackson.core.*;
+import com.fasterxml.jackson.core.PrettyPrinter;
```
cowtowncoder: Please do not reformat code in places where you don't change it.

src/main/java/com/fasterxml/jackson/dataformat/xml/ser/XmlSerializerProvider.java
cowtowncoder: A brief comment would be appreciated... however, are you sure this is safe? Keep in mind that it extends `DefaultSerializerProvider` so I think it would be better to create a copy (super-class single-arg constructor to use).

src/main/java/com/fasterxml/jackson/dataformat/xml/XmlMapper.java
gtrog: Sorry about that, IntelliJ auto-formatted it, I'll revert those

cowtowncoder:
Excellent, thank you! Will merge.

Issue #238, opened 2017-05-29
Error:trouble processing "javax/xml/bind/Binder.class"


boomsya:
android studio 2.3.2
dependencies for this lib:
compile group: 'com.fasterxml.jackson.dataformat', name: 'jackson-dataformat-xml', version: '2.9.0.pr3'
compile group: 'javax.xml.bind', name: 'jaxb-api', version: '2.2.12'

Error:Execution failed for task ':app:transformClassesWithDexForDevDebug'.
> com.android.build.api.transform.TransformException: java.lang.RuntimeException: java.lang.RuntimeException: Unable to pre-dex 'C:\Users\kolio\.gradle\caches\modules-2\files-2.1\javax.xml.bind\jaxb-api\2.2.12\4c83805595b15acf41d71d49e3add7c0e85baaed\jaxb-api-2.2.12.jar' to 'Z:\build\intermediates\transforms\dex\dev\debug\folders\1000\10\jaxb-api-2.2.12_3e8723a2ab4f6ea021e7276f169b0687fdc79ad6'

cowtowncoder:
This sounds like Android build problem, so I can't help much.
You may want to ask on mailing lists if other Android developers might know what gives.

Since gradle is involved perhaps you can also clear modules it has cached; at least Maven sometimes manages to get into state where cached artifacts are corrupt locally. So perhaps something similar could occur with gradle.

fan90910:
i met the same question, i found the reason is android studio project's target jdk version is jdk1.7, but there are some files conflicted with files in stax-api-1.0-2.jar, such as file 'javax/xml/stream/EventFilter.class'. So, anyone has some advices?

fan90910:
@boomsya @cowtowncoder

cowtowncoder:
@fan90910 those classes form Stax API which is include in JDK 6 and above, but formerly were excluded from Android. If they are now part of SDK now just exclude `stax-api-1.0-2.jar` and rely on SDK provided ones. They are just API types (interfaces, abstract classes), not implementations.

Issue #142, opened 2015-03-23
Problems with stax-api dependency on Android


cowtowncoder:
(note: offshoot of #116)

Although JDK 1.6 does include Stax API (`javax.xml.stream`), Google has chosen to create an arbitrary subset of the same and is not included this API. Earlier it even blocked attempts at including any non-included dependencies in `javax.` namespace, but apparently this has been resolved, at least as far as Stax API is included (by whitelisting?).

But. Since dependency is needed on Android, there is need to either add an otherwise unnecessary dependency (unnecesasry for J2SE/J2EE), which may add unnecessary jar in builds on non-Android platforms; or to add instructions to help Android developers add the dependency explicitly.

I will first add a note on README, for approach two.

keshav-iitr:
@cowtowncoder Do you have any solution for this issue?  I am also facing this issue please help.

cowtowncoder:
@keshav-iitr I don't know anything more; but mailing list (`jackson-user`) has more Android developers; maybe someone knows.

cowtowncoder:
I don't know what else could or should be done here -- the fact remains that `stax-api` appears necessary for Android; but since it is provided by all J2SE platforms current versions of Jackson run (Java 6 and above), it does not make sense to add as a hard dependency.

cternes:
I get the following error while trying to include the Stax API explicitly on Android (v5.0):

> Ill-advised or mistaken usage of a core class (java.* or javax.*) when not building a core library.
> 
> This is often due to inadvertently including a core library file in your application's project, when using an IDE (such as Eclipse). If you are sure you're not intentionally defining a core class, then this is the most likely explanation of what's going on.
> 
> However, you might actually be trying to define a class in a core namespace, the source of which you may have taken, for example, from a non-Android virtual machine project. This will most assuredly not work. At a minimum, it jeopardizes the compatibility of your app with future versions of the platform.
> It is also often of questionable legality.
> 
> If you really intend to build a core library -- which is only appropriate as part of creating a full virtual machine distribution, as opposed to compiling an application -- then use the "--core-library" option to suppress this error message.
> 
> If you go ahead and use "--core-library" but are in fact building an application, then be forewarned that your application will still fail to build or run, at some point. Please be prepared for angry customers who find, for example, that your application ceases to function once they upgrade their operating system. You will be to blame for this problem.
> 
[...]

PR #280, opened 2018-02-07
Amending Android usage notes.

roberttaylor426:
As of recent versions of Android the current Android-specific instructions in the README are broken. 

If you declare an explicit dependency on the Stax API the Android build tools will fail your build when you attempt to package a non-debuggable version of your app (necessary for publishing to the Play Store). This issue was reported [here](https://github.com/FasterXML/jackson-dataformat-xml/issues/142) and [here](https://github.com/FasterXML/jackson-dataformat-xml/issues/238).

There are (extremely hacky) [workarounds](https://stackoverflow.com/questions/31360025/using-jackson-dataformat-xml-on-android) that exist, but others have reported unresolved problems with them.

I would advocate explicitly declaring that Android usage is currently unsupported since:
 * In the comments to the aforementioned (closed) issues the official response reads something like: `I don't know: I don't develop on Android. Please ask on mailing lists.`
 * Based on the assumption that it was supported, I personally invested a fair amount of time adopting your library in my Android project, only to find when I came to publish my app to the Play Store that it wasn't. It seems like it would be fairer to be brutally honest upfront with your users

Thanks otherwise for a nice library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,24 +77,9 @@ XmlMapper xmlMapper = new XmlMapper(module);
 as many features that `XmlMapper` needs are provided by `JacksonXmlModule`; default
 `XmlMapper` simply constructs module with default settings.
 
-## Android quirks
+## Android
 
-While usage on Android is the same as on standard JDKs, there is one thing that may cause issues:
-since Google has chosen not to include whole JDK 1.6 API, `Stax` API (package `javax.xml.stream`) is missing.
-This means that one has to add dependency explicitly. With Maven it can be done with
-
-```xml
-<dependency>
-    <groupId>javax.xml.stream</groupId>
-    <artifactId>stax-api</artifactId>
-    <version>1.0-2</version>
-    <scope>compile</scope>
-</dependency>
-```
-
-or, if using other build tools, include similar dependency or download actual jar from
-
-    http://repo1.maven.org/maven2/javax/xml/stream/stax-api/1.0-2/
+Usage of this library on Android is currently not supported. This is due to the fact that the Stax API is unavailable on the Android platform, and attempts to declare an explicit dependency on the Stax API library will result in errors at build time (since the inclusion of the `javax.*` namespace in apps is restricted).
 
 ## Serializing POJOs as XML
 


cowtowncoder:
@roberttaylor426 Fair enough. I just can't believe how bone-headed Google can be, in their badly butchered subsetting of JDK APIs. :-( :-( :-(
I had naively assumed that after years and years of progress, such a simple improvement as either including proper XML API subset of Java 6, OR allowing white-listing, this would be non-issue for newer SDKs. Apparently not.

About the only thing I would ask is that if you know of good explanation of underlying problem (yes, I understand that you simply can not get those classes to load), wrt Stax API, that would still be good to link. I would think that outside of Jackson, attempts have been made to use JAXB, which should face similar issues (it does not require use of Stax, SAX works, but API itself exposes StaxSource.
I guess that's what's alluded on

https://stackoverflow.com/questions/5461127/using-jaxb-with-google-android

as well...

roberttaylor426:
Honestly it's quite hard to find an authoritative link that explains the problem.

The absence of required libraries is certainly evident in the [Android Javadocs](https://developer.android.com/reference/javax/xml/package-summary.html).

The fact that Android prohibits the inclusion of code inhabiting the "javax.*" namespace is fairly well detailed around the web, just not from Google themselves. [This link](https://www.docx4java.org/blog/2012/05/jaxb-can-be-made-to-run-on-android/) does a pretty good job of explaining the problem and providing a potential workaround, as does the one I [mentioned above](https://stackoverflow.com/questions/31360025/using-jackson-dataformat-xml-on-android).

Do you feel any of these links would be appropriate to add to the README?

cowtowncoder:
@roberttaylor426 yes, I agree, knowledge exists, just in fragments. I think I'll merge this in and add links. I actually worked quite a bit with Sean B few years back on this module (mostly wrt trying to figure out excessive startup time spend, compared to json backend), and this is why I was hoping situation from Android end would have improved wrt SDK.

Thank you for your help with documentation & apologies of for the mess (even if there isn't much I feel I can do aside from adding notes about situation).

PR #337, opened 2019-04-08
Fix #336

### LINKED ISSUE #336: WRITE_BIGDECIMAL_AS_PLAIN Not Used When Writing Pretty
When a xml pretty printer is used, the flag of WRITE_BIGDECIMAL_AS_PLAIN is not taken into account for that if statement.

Is there a reason for this?  If not, I can see if I can create a PR to resolve.

https://github.com/FasterXML/jackson-dataformat-xml/blob/master/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java#L1126

kevindaub:
Fixed usage of WRITE_BIGDECIMAL_AS_PLAIN when XML pretty printer is
configured.


cowtowncoder:
Thanks!

PR #357, opened 2019-08-17
Fix for #336, multiple class adapted for the feature.

perilbrain:
- Accepts new annotation for collection type.

Hey I have made a few improvements to fix the issue. All tests cases are passing but I definitely need a review for this commit.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/XmlAnnotationIntrospector.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/XmlAnnotationIntrospector.java
@@ -2,7 +2,9 @@ package com.fasterxml.jackson.dataformat.xml;
 
 import com.fasterxml.jackson.databind.AnnotationIntrospector;
 import com.fasterxml.jackson.databind.introspect.Annotated;
+import com.fasterxml.jackson.databind.introspect.AnnotatedClass;
 import com.fasterxml.jackson.databind.introspect.AnnotationIntrospectorPair;
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement;
 import com.fasterxml.jackson.module.jaxb.JaxbAnnotationIntrospector;
 
 /**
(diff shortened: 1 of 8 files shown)


src/main/java/com/fasterxml/jackson/dataformat/xml/ser/XmlSerializerProvider.java
```diff
@@ -107,7 +109,8 @@ public void serializeValue(JsonGenerator gen, Object value, JavaType rootType,
             _initWithRootName(xgen, rootName);
             asArray = TypeUtil.isIndexedType(rootType);
             if (asArray) {
-                _startRootArray(xgen, rootName);
+                String indexedRootName = _rootNameLookup.findWrapperForIndexedType(getTypeOfCollection(rootType), _config);
```
perilbrain: // May be `value` is right candidate as tested in spring with v2.9.9
String indexedRootName = _rootNameLookup.findWrapperForIndexedType(getTypeOfCollection(value), _config);

cowtowncoder:
Ah, this looks good I think; I will try to go through with it in more detail, but wanted to mention that this is on my radar for things to go in 2.10 (but I'm also bit overload with all changes :) ).

Only one more question: have I already asked for CLA? If not, it's here:

and we'd need a filled copy before first merged contribution (apologies if I have already asked and gotten one).
Easiest way usually is to print it, fill & sign, scan, email to `info` at fasterxml dot com.

src/main/java/com/fasterxml/jackson/dataformat/xml/XmlAnnotationIntrospector.java
```diff
@@ -183,5 +196,10 @@ public Boolean isOutputAsCData(Annotated ann) {
         public void setDefaultUseWrapper(boolean b) {
             // not used with JAXB
         }
+
+        @Override
+        public String getWrapperForIndexedType(AnnotatedClass ac) {
+            return JacksonXmlRootElement.DEFAULT_WRAPPER_NAME;
```
cowtowncoder: I think I would rather return empty String, handle defaulting outside of this method; otherwise AI chaining will not work properly (as it is not possible to distinguish default value from explicit "item").

PR #366, opened 2019-10-23
XML containing xsi:nil is improperly parsed

lkorth:
XML containing a nested field with `xsi:nil="true"` causes following fields to be deserialized at the wrong hierarchy. I've added a unit test to demonstrate this incorrect behavior.

This appears to be a new bug in 2.10.0 as a result of https://github.com/FasterXML/jackson-dataformat-xml/commit/4938036087bed3ebdfaf6c98db14d269179c30a8

### DIFF
--- a/src/test/java/com/fasterxml/jackson/dataformat/xml/deser/XsiNil354Test.java
+++ b/src/test/java/com/fasterxml/jackson/dataformat/xml/deser/XsiNil354Test.java
@@ -14,6 +14,20 @@ public class XsiNil354Test extends XmlTestBase
         }
     }
 
+    protected static class Parent {
+        public Level1 level1;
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Thank you for reporting this -- I'll merge the test, and refactor it to go under `failing`.
I don't know when I'll have time to look but it'd be good to straighten out aspects of `xsi:nil` (and related xsi attributes) for sure.

lkorth:
Thanks! Sorry, forgot to set the right base branch on the PR, it was supposed to be 2.11

cowtowncoder:
That's fine: I actually brought it back to 2.10 since it'd be good to fix there, not just 2.11 (although we'll see)

PR #407, opened 2020-05-23
Migrate to Jakarta EE 9

tamersaadeh:
Jakarta EE 9 platform renames all the "old" namespaces from javax.xml.bind.* to jakarta.xml.bind.*. This pull request performs this migration so that when Jacakrta EE 9 is released in [June 2020](https://eclipse-ee4j.github.io/jakartaee-platform/jakartaee9/JakartaEE9ReleasePlan#proposed-progress-reviews) it will be compatible.

I ran mvn test and [travis builds](https://travis-ci.com/github/tamersaadeh/jackson-dataformat-xml/builds/167887594), with a single failure that does not seem related to the changes.

Note 1: this is completely incompatible with the old namespace, so probably the major version of the library should be bumped.
Note 2: currently depends on pre-released versions of the JAXB specification (RC3). This can be updated when the final version is released.
Note 3: this PR is similar to my PR for FasterXML/jackson-modules-base#95


cowtowncoder:
Note for posterity: only changes test classes so should be safe. Also looks like `3.0.0` was just released.

cowtowncoder:
Ok: I think this makes sense now that Jackson 2.13 adds a new set of modules for new Jakarta XML Bind vs old Javax/JAXB.
Will have to massage PR a bit but should be able to make it work.

Thank you @tamersaadeh!

Issue #405, opened 2020-05-22
Mixed content not exposed through `FromXmlParser`, lost by `JsonNode`


cowtowncoder:
Currently (2.11) "Mixed content" is only exposed in case of textual content preceding an element, like:

```xml
<root>
   some text
  <child> ... </child>
</root>
```

but is not retained or exposed in token stream if it comes after a close element, so none of text segments in:

```xml
<root>
  <child>...</child>
  first text
  <child>...</child>
  last text
</root>
```

are exposed. This means that although as per #403 `JsonNode` could now express such content (considering nominal key of "" for text segements), but does not get them as `FromXmlParser` does not create matching logical tokens.
This should be fixed.

cowtowncoder:
Implemented for 2.12(.0). Couple of limitations:

1. While ordering of text segments will be retained (wrt each other), since all will be under logical key of ""  (empty String), total ordering for all segments is not retained
2. Blank (all-whitespace) text segments are not retained (not even between end-start-element) -- currently they must be skipped to make indentation work for regular content.

So it may make sense to introduce something more specific for really accurate XML content (besides above-mentioned issues, element/attribute distinction is not preserved either).
Choices include:

* Make DOM `Element` bind directly from underlying XML elements (instead of trying to parse textual values
* Create something similar to DOM (JDOM, DOM4j); `XmlNode` for example, make that bind.

But work for those need to go under different issue.

PR #436, opened 2020-11-30
Small update of the known limitations section of the README

PascalSchumacher:
By the way: Thank you for providing Jackson. 👍


cowtowncoder:
Thank you @PascalSchumacher! Glad to hear you like Jackson, and updates to documentation are very much appreciated.

I also noticed that the adjacent not on Mixed Content is also outdated now, see #405 -- not full solution, but at least text segments are exposed with virtual key "" (empty String) by parser, and can be bound to some types.

PR #456, opened 2021-03-10
Fix JsonAlias with unwrapped lists

westinrm:
WrapperHandlingDeserializer does not add aliases for unwrapped properties to the set of unwrapped names,
causing deserialization of unwrapped lists using aliases to fail.

This has the slightly strange behavior that if you mix aliases you'll end up with the latter sequence.
Ex `<val>1</val><alias>2</alias>` will deserialize to `[2]` instead of `[1,2]`

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/deser/WrapperHandlingDeserializer.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/deser/WrapperHandlingDeserializer.java
@@ -92,6 +92,9 @@ public class WrapperHandlingDeserializer
             }
             // not optimal; should be able to use PropertyName...
             unwrappedNames.add(prop.getName());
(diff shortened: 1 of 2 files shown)


westinrm:
Sorry for the long delay here -- I've sent a CLA to you that should cover any contributions by me.

cowtowncoder:
@westinrm Np. I don't yet see it.. but will keep checking, might be just delay.

westinrm:
It was sent on Friday, Mar 19th around 11pm GMT.

cowtowncoder:
uff. Of course it went to "spam" for some reason. Ok, mystery solved.

cowtowncoder:
Thank you again! Merged in 2.11 branch, but most likely to be available in 2.12.3 next (2.11 may or may not get another minor version release; if one is done it'll be 2.11.5)

Issue #490, opened 2021-08-25
Problem when using defaultUseWrapper(false) in combination with polymorphic types


daniel-kr:
Hello,

we are currently trying to upgrade `jackson-dataformat-xml` in our product from `2.11.4` to `2.12.4`.
However, with the new version we encounter problems when it comes to using `defaultUseWrapper(false)` in combination with polymorphic types.

Given the following interface and concrete implementation:

```java
@JsonTypeInfo(
        use = JsonTypeInfo.Id.NAME,
        property = "type")
@JsonSubTypes({
        @JsonSubTypes.Type(value = MyType.class, name = "myType"),
})
public interface IMyType {
}
```

```java
public class MyType implements IMyType {

    @JsonCreator
    public MyType(
            @JsonProperty("stringValue") String stringValue,
            @JsonProperty("typeNames") Collection<String> typeNames) {
        this.stringValue = stringValue;
        this.typeNames = typeNames;
    }

    public final String stringValue;
    public final Collection<String> typeNames;

    // equals() and hashcode() removed from the snippet
}
```

Now doing a roundtrip serialization using the following code:

```java
public class Main {

    public static void main(String[] args) throws JsonProcessingException {
        XmlMapper xmlMapper = XmlMapper.builder()
                //.configure(FromXmlParser.Feature.EMPTY_ELEMENT_AS_NULL, true)
                //.configure(DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY, true)
                .defaultUseWrapper(false).build();

        MyType myType = new MyType("hello", List.of("type1", "type2"));

        String stringValue = xmlMapper.writeValueAsString(myType);
        System.out.println(stringValue);
        IMyType outputType = xmlMapper.readValue(stringValue, IMyType.class);
        Preconditions.checkState(myType.equals(outputType));
    }
}
```

While this works great on version `2.11.4`, it produces the following exception during de-serialization on `2.12.4`:

```
com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot construct instance of `java.util.ArrayList` (although at least one Creator exists): no String-argument constructor/factory method to de-serialize from String value ('type1')
```

The produced XML looks as follows on both versions:

```xml
<MyType type="myType">
    <stringValue>hello</stringValue>
    <typeNames>type1</typeNames>
    <typeNames>type2</typeNames>
</MyType>
```

During the tests we recognized the following behavior:

[...]

PR #497, opened 2021-10-13
Fix: Problem when using defaultUseWrapper(false) in combination with polymorphic types

lglauer:
This PR fixes a regression introduced by https://github.com/FasterXML/jackson-dataformat-xml/commit/d9738d92bfe1ea4ff59c339cc42348fb70936ae4 , which lead to [this bug report](https://github.com/FasterXML/jackson-dataformat-xml/issues/490).


cowtowncoder:
Thank you @lglauer ! I think this seems like a reasonable take: I wish I had time to go through it thoroughly (just wrt possibility of off-by-one for nesting); but since there are no regressions, and since I know what causes "missing" `START_OBJECT`, I'll use this.
I did add change as part of commit with test case so will close this PR, but noted you as the contributor of the fix in release notes since this is where it came from.

I'll include it in 2.12.6 release, hopefully due out soon.

It also seems like there are some further issues wrt #490, will note more there.

PR #504, opened 2021-11-23
fix a flaky test

arianacai1997:
`testNoElementWrapper` is flaky since the order of names is nondeterministic. I fixed it by adding all possibilities of the orders.


cowtowncoder:
Looks good, thank you!

PR #527, opened 2022-05-25
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+      interval: "weekly"


cowtowncoder:
Could you please link to some documentation that talks a little more about it?
I have had pretty bad experiences with Dependabot (it has been net negative adding little value and lots of noises) so I am bit hesitant to add anything before fully understanding how it works. It does sound like this _could_ be useful addition, don't get me wrong, but... once bitten twice shy. :)

naveensrinivasan:
> Could you please link to some documentation that talks a little more about it?
> I have had pretty bad experiences with Dependabot (it has been net negative adding little value and lots of noises) so I am bit hesitant to add anything before fully understanding how it works. It does sound like this _could_ be useful addition, don't get me wrong, but... once bitten twice shy. :)

I have provided documentation links in the PR as well as in the commit. Here are some more documentation https://docs.github.com/en/code-security/dependabot/working-with-dependabot/troubleshooting-the-detection-of-vulnerable-dependencies

Please let me know what else you are looking for. 

This is only for GH Actions.

cowtowncoder:
Ok let's see if this turns out to be useful -- can always turn off if there is too much noise.

PR #534, opened 2022-06-25
add tests for doubles/floats

pjfanning:
I was having a look to see if the code in this module needed to change to best support the fast double parser in jackson-core and saw that there weren't many tests for doubles/floats.

Java 17 CI build fails but that looks like a pre-existing issue.


cowtowncoder:
Sounds good, thank you!

PR #543, opened 2022-08-26
QName Deserializer

vitorpamplona:
QName is a special case of an XML Deserializer. This PR parses the contents of the QName and assigns the correct namespace from the XML stack. This mimics the behavior of JAXB parsers. 

An XML like this: 

```xml
<parent xmlns:t="urn:example:types:r1">
    <level1 name="t:DateTime" />"
</parent>";
```

Should automatically generate a `parent.level1.name` as `QName("urn:example:types:r1", "DateTime", "t")`

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/deser/QNameDeserializer.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/deser/QNameDeserializer.java
@@ -1,20 +1,23 @@
-package tools.jackson.dataformat.xml.deser;
+package com.fasterxml.jackson.dataformat.xml.deser;
 
 import javax.xml.namespace.QName;
 
-import tools.jackson.core.JacksonException;
-import tools.jackson.core.JsonParser;
-import tools.jackson.databind.DeserializationContext;
-import tools.jackson.databind.ValueDeserializer;
+import java.io.IOException;
 
-public class QNameDeserializer extends ValueDeserializer {
-    ValueDeserializer<?> originalDeserializer;
-    public QNameDeserializer(ValueDeserializer<?> deserializer) {
+import com.fasterxml.jackson.core.JacksonException;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+
+
+public class QNameDeserializer extends JsonDeserializer {
+    JsonDeserializer<?> originalDeserializer;
+    public QNameDeserializer(JsonDeserializer<?> deserializer) {
         originalDeserializer = deserializer;
     }
 
     @Override
-    public Object deserialize(JsonParser p, DeserializationContext ctxt) throws JacksonException {
+    public Object deserialize(JsonParser p, DeserializationContext ctxt) throws IOException, JacksonException {
         QName qName = (QName) originalDeserializer.deserialize(p, ctxt);
 
         if (qName.getLocalPart().indexOf(":") > 0) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/fasterxml/jackson/dataformat/xml/deser/QNameDeserializer.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    @Override
+    public Object deserialize(JsonParser p, DeserializationContext ctxt) throws IOException, JacksonException {
+        QName qName = (QName) originalDeserializer.deserialize(p, ctxt);
+
+        if (qName.getLocalPart().indexOf(":") > 0) {
+            String prefix = qName.getLocalPart().split(":")[0];
+            String localPart = qName.getLocalPart().split(":")[1];
+            String namespace = ((FromXmlParser)ctxt.getParser()).getStaxReader().getNamespaceContext().getNamespaceURI(prefix);
```
cowtowncoder, line 26: This won't work if content is buffered, so I am not sure this approach is valid unfortunately.

cowtowncoder:
Ok, first of all: thank you for contributing this patch!
I can see what it is attempting to do and that makes sense.

Unfortunately I am not sure this can be implemented in robust manner: the main problem being that there is no guarantee that

1. There is a Stax parser associated with content -- in particular when buffering (using `TokenBuffer`) is used (most commonly when `@JsonCreator` annotated constructor used)
2. Parser itself may not point to actual XML content element (in case of textual content it's probably pointing to closing element, in case of elements)

latter might not be a huge issue as long as namespace resolution context is still available, but former is problematic.
I guess one could make code check accessibility and avoid lookup if parser not available. But that would be hugely confusing since it "sometimes works, sometimes not" (with no obvious signs to user).

src/main/java/com/fasterxml/jackson/dataformat/xml/deser/QNameDeserializer.java
vitorpamplona: Humm.. interesting.. do you have a simple test case for the Buffered use case? 

I am using the same design in our solution so, even if it is not a viable solution here, I would love to improve the code in our solution. Or at least know em when exactly it will fail.

cowtowncoder: It's a side-effect of processing, not a feature to enable: if you create a POJO with `@JsonCreator(mode = JsonCreator.Mode.PROPERTIES)`, property values will likely be buffered. Similarly for polymorphic types if Type Id is not the first value deserialized.
Basically any case where stream-order is not the same as the order in which values are needed will be buffered.

vitorpamplona: So, the correct way to solve this would be to change the token streams to include namespaces and prefixes available at each one of them?

cowtowncoder: Yes. `TokenBuffer`, however, is not XML specific and cannot (should not) be changed.
2.13 allows replacement of buffer implementation so we could have `XmlTokenBuffer` sub-class (or such) -- XML backend was the reason to add this -- but there's then the question of how to access information. Probably `FromXmlParser` and `XmlTokenBuffer` could implement an interface (to be added) for additional accessors.

I have not, unfortunately, had time to take this approach any further but this would be a very valuable improvement and unblock work on solving multiple issues.
So I would be happy to help you or anybody who had time to look into improvements in this area.

Issue #546, opened 2022-09-18
XmlMapper is adding a new empty line at the end of file


M4NN3:
XmlMapper, when using `writerWithDefaultPrettyPrinter `and writing value to a new fle is adding an empty line at the end of the file.
This is not happening when using `writeValueAsString`

Roomka:
I took a look at the source code and it looks like this behavior is caused by https://github.com/FasterXML/jackson-dataformat-xml/blob/2.14/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java#L619. 

From my understanding, the intention was to add a new line feed to separate entities outside of the root element (that is, in prolog or epilog), but it is also added after the document body in the scenario in which a document doesn't contain an epilog. 

I am not sure if it was intended to behave in this way or if it can be considered a bug.

cowtowncoder:
Timing now is pretty tight -- I am about to publish 2.14.0 final in 2-3 days -- but if anyone could provide a test to show that with change to remove extra linefeed:

1. Case WITH epilog (comment or PI) works as expected (comment/PI is on line following last closing element) AND
2. Case WITHOUT anything else works (only single linefeed after last closing element)

I'd be happy to merge said PR. I don't have time to do this myself right now but would be happy to code review, get merged.

Roomka:
Hi @cowtowncoder,

not sure if I am missing something but from my understanding, there is no way to add the epilog in an object mapped by the object mapper. 

I opened a PR https://github.com/FasterXML/jackson-dataformat-xml/pull/554 removing the part of the code that adds extra linefeed and updating the unit test affected by that change accordingly.

cowtowncoder:
Ah. Yes, you are right, there is no way to add prolog or epilog stuff quite yet. Thank you for #554 -- I'll try to have a look soon, but am just bit swamped at work. But I'll get there eventually. :)

cowtowncoder:
Ok as far as I can see this is actually behaving as I would expect.

To change behavior what you can do is to sub-=class `DefaultXmlPrettyPrinter` and override method to do nothing:

    public void writePrologLinefeed(XMLStreamWriter2 sw) throws XMLStreamException { }

This would change serialization so that no linefeeds would be added before opening root element or after closing one.

In theory it could be possible to make this configurable but it would get bit tricky, esp. if was to go through `XmlMapper`.

[...]

PR #554, opened 2022-11-09
Removed prolog line feed in XMLPrettyPrinter as the current version m…

Roomka:
Removed prolog line feed in XMLPrettyPrinter as the current version mapper doesn't support prolog object mapping


cowtowncoder:
Oh. This is odd -- from the original description I got the impression there was an EXTRA linefeed.
But this seems to just the usual single linefeed at the end of the document, which makes sense for indented case (and would not be added without pretty-printing).

So not 100% sure this actually makes sense... will add a note on #546.

cowtowncoder:
I don't think this makes sense any more, closing.

PR #562, opened 2022-12-12
Add NumberDeserWithXMLTests to show num len check works

pjfanning:
Check that the XmlFactory streamReadConstraints are used properly


cowtowncoder (approved):

cowtowncoder:
Sounds good. I'll change implementation slightly in 2.15 but the basic idea is the same.

cowtowncoder:
I think we may have similar issue with textual format backends too (CSV and Properties; YAML should be fine).

pjfanning:
> I think we may have similar issue with textual format backends too (CSV and Properties; YAML should be fine).

I have a PR that adds more tests to dataformats-text

Issue #560, opened 2022-12-08
Add `DefaultXmlPrettyPrinter.withCustomNewLine()` to configure linefeed for XML pretty-printing


koalalam:
Jackson version: 2.14.1

In `ObjectMapper` case, we can do the following:

```java
objectMapper.setDefaultPrettyPrinter(new DefaultPrettyPrinter().withObjectIndenter(new DefaultIndenter().withLinefeed("\n")));
```

However in `XmlMapper`. the default implementation [Lf2SpacesIndenter](https://github.com/FasterXML/jackson-dataformat-xml/blob/jackson-dataformat-xml-2.14.1/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java#L504)
1) Is protected
2) Doesn't seem to support a custom line separator

Does it mean we need to provide our own implementation of `Indenter` just to have a different line separator?

cowtowncoder:
Seems like an oversight, thank you for reporting it.

If you have time and interest, a PR would make sense, if you end up trying writing a custom implementation anyway.

PR #568, opened 2023-02-05
Feature : Allow setting custom linefeed on DefaultXmlPrettyPrinter 

JooHyukKim:
- for #560 

BEFORE
- Linefeed is set when `static { }` block from `Lf2SpacesIndenter` is run.
- both `DefaultXmlPrettyPrinter`and `Indenters` both use static `Lf2SpacesIndenter.SYSTEM_LINE_SEPARATOR` 

AFTER
- `_lineFeed` is set when `static { }` block from `DefaultXmlPrettyPrinter` is run.
- `Indenters` ONLY customize indentations and linefeed value is given by `DefaultXmlPrettyPrinter`

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
@@ -82,7 +82,7 @@ public class DefaultXmlPrettyPrinter
      * Can later set custom new line with withCustomNewLine method.
      * @since 2.15
      */
-    private final static String SYSTEM_DEFAULT_NEW_LINE;
+    private static final String SYSTEM_DEFAULT_NEW_LINE;
     static {
         String lf = null;
         try {
@@ -90,8 +90,13 @@ public class DefaultXmlPrettyPrinter
         } catch (Exception t) { } // access exception?
         SYSTEM_DEFAULT_NEW_LINE = lf;
     }
-    
     protected String _newLine = SYSTEM_DEFAULT_NEW_LINE;
+
+    static final int SPACE_COUNT = 64;
+    static final char[] SPACES = new char[SPACE_COUNT];
+    static {
+        Arrays.fill(SPACES, ' ');
+    }
 
     /*
     /**********************************************************
@@ -143,7 +148,7 @@ public class DefaultXmlPrettyPrinter
     public void spacesInObjectEntries(boolean b) { _spacesInObjectEntries = b; }
 
     /**
-     * Sets custom linefeed
+     * Sets custom new-line
      * @since 2.15
      */
     public DefaultXmlPrettyPrinter withCustomNewLine(String newLine) {
@@ -513,7 +518,7 @@ public class DefaultXmlPrettyPrinter
         {
             sw.writeRaw(" ");
         }
-        
+
         @Override
         public void writeIndentation(JsonGenerator g, int level) throws IOException
         {
@@ -532,13 +537,6 @@ public class DefaultXmlPrettyPrinter
         implements Indenter, java.io.Serializable
     {
         private static final long serialVersionUID = 1L;
-
-        final static int SPACE_COUNT = 64;
-        final char[] SPACES = new char[SPACE_COUNT];
-
-        public Lf2SpacesIndenter() {
-            Arrays.fill(SPACES, ' ');
-        }
 
         @Override
         public boolean isInline() { return false; }
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
@@ -96,7 +96,7 @@ public class DefaultXmlPrettyPrinter
         }
     }
     
-    protected static String _lineFeed = DEFAULT_LINE_FEED;
+    protected String _lineFeed = DEFAULT_LINE_FEED;
 
     /*
     /**********************************************************
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
@@ -148,13 +148,11 @@ public class DefaultXmlPrettyPrinter
     public void spacesInObjectEntries(boolean b) { _spacesInObjectEntries = b; }
 
     /**
-     * Sets custom new-line
+     * Sets custom new-line.
      * @since 2.15
      */
     public DefaultXmlPrettyPrinter withCustomNewLine(String newLine) {
-        if (newLine != null) {
-            _newLine = newLine;
-        }
+        _newLine = newLine != null ? newLine : SYSTEM_DEFAULT_NEW_LINE;
         return this;
     }
 
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
@@ -133,6 +133,7 @@ public class DefaultXmlPrettyPrinter
         _objectIndenter = base._objectIndenter;
         _spacesInObjectEntries = base._spacesInObjectEntries;
         _nesting = base._nesting;
+        _newLine = base._newLine;
     }
 
     public void indentArraysWith(Indenter i)
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
@@ -78,8 +78,8 @@ public class DefaultXmlPrettyPrinter
     protected boolean _spacesInObjectEntries = true;
 
     /**
-     * By default, will try to set as System.getProperty("line.separator")
-     * Can later set custom lineFeed with withCustomlineFeed method
+     * By default, will try to set as System.getProperty("line.separator").
+     * Can later set custom lineFeed with withCustomlineFeed method.
      * @since 2.15
      */
     private final static String DEFAULT_LINE_FEED;
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
@@ -87,7 +87,7 @@ public class DefaultXmlPrettyPrinter
         String lf = null;
         try {
             lf = System.getProperty("line.separator");
-        } catch (Throwable t) { } // access exception?
+        } catch (Exception t) { } // access exception?
         SYSTEM_DEFAULT_NEW_LINE = lf;
     }
     
@@ -147,7 +147,6 @@ public class DefaultXmlPrettyPrinter
      * @since 2.15
      */
     public DefaultXmlPrettyPrinter withCustomNewLine(String newLine) {
-        // 06-Feb-2023, joohyukkim: when JacksonException extends RuntimeExceptions, throw it?
         if (newLine != null) {
             _newLine = newLine;
         }
@@ -485,7 +484,7 @@ public class DefaultXmlPrettyPrinter
     /**
      * Dummy implementation that adds no indentation whatsoever
      */
-    protected class NopIndenter
+    protected static class NopIndenter
         implements Indenter, java.io.Serializable
     {
         private static final long serialVersionUID = 1L;
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
@@ -66,7 +66,7 @@ public class DefaultXmlPrettyPrinter
      * system-specific linefeeds, and 2 spaces per level (as opposed to,
      * say, single tabs)
      */
-    protected Indenter _objectIndenter = new Lf2SpacesIndenter();
+    protected Indenter _objectIndenter = new Lf2SpacesIndenter(this);
 
     // // // Config, other white space configuration
 
@@ -459,6 +459,11 @@ public class DefaultXmlPrettyPrinter
         _justHadStartElement = false;
     }
 
+    @Override
+    public String lineFeed() {
+        return _lineFeed;
+    }
+
     // @since 2.12
     public void writeLeafXsiNilElement(XMLStreamWriter2 sw,
             String nsURI, String localName)
@@ -545,7 +550,11 @@ public class DefaultXmlPrettyPrinter
             Arrays.fill(SPACES, ' ');
         }
 
-        public Lf2SpacesIndenter() { }
+        private final XmlPrettyPrinter xmlPrettyPrinter;
+
+        public Lf2SpacesIndenter(XmlPrettyPrinter xmlPrettyPrinter) {
+            this.xmlPrettyPrinter = xmlPrettyPrinter;
+        }
 
         @Override
         public boolean isInline() { return false; }
@@ -553,7 +562,7 @@ public class DefaultXmlPrettyPrinter
         @Override
         public void writeIndentation(XMLStreamWriter2 sw, int level) throws XMLStreamException
         {
-            sw.writeRaw(_lineFeed);
+            sw.writeRaw(xmlPrettyPrinter.lineFeed());
             level += level; // 2 spaces per level
             while (level > SPACE_COUNT) { // should never happen but...
             	sw.writeRaw(SPACES, 0, SPACE_COUNT); 
@@ -565,7 +574,7 @@ public class DefaultXmlPrettyPrinter
         @Override
         public void writeIndentation(JsonGenerator jg, int level) throws IOException
         {
-            jg.writeRaw(_lineFeed);
+            jg.writeRaw(xmlPrettyPrinter.lineFeed());
             level += level; // 2 spaces per level
             while (level > SPACE_COUNT) { // should never happen but...
                 jg.writeRaw(SPACES, 0, SPACE_COUNT); 
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
@@ -149,7 +149,7 @@ public class DefaultXmlPrettyPrinter
      * Sets custom linefeed
      * @since 2.15
      */
-    public DefaultXmlPrettyPrinter withCustomlineFeed(String lineFeed) {
+    public DefaultXmlPrettyPrinter withCustomLineFeed(String lineFeed) {
         // 06-Feb-2023, joohyukkim: when JacksonException extends RuntimeExceptions, throw it?
         if (lineFeed != null) {
             _lineFeed = lineFeed;
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
@@ -538,6 +538,8 @@ public class DefaultXmlPrettyPrinter
     {
         private static final long serialVersionUID = 1L;
 
+        public Lf2SpacesIndenter() { }
+
         @Override
         public boolean isInline() { return false; }
 
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
@@ -66,7 +66,7 @@ public class DefaultXmlPrettyPrinter
      * system-specific linefeeds, and 2 spaces per level (as opposed to,
      * say, single tabs)
      */
-    protected Indenter _objectIndenter = new Lf2SpacesIndenter(this);
+    protected Indenter _objectIndenter = new Lf2SpacesIndenter();
 
     // // // Config, other white space configuration
 
@@ -88,12 +88,7 @@ public class DefaultXmlPrettyPrinter
         try {
             lf = System.getProperty("line.separator");
         } catch (Throwable t) { } // access exception?
-
-        if (lf != null) {
-            DEFAULT_LINE_FEED = lf;
-        } else {
-            DEFAULT_LINE_FEED = "\n"; // incase system has changed property name? line.separator
-        }
+        DEFAULT_LINE_FEED = lf;
     }
     
     protected String _lineFeed = DEFAULT_LINE_FEED;
@@ -459,11 +454,6 @@ public class DefaultXmlPrettyPrinter
         _justHadStartElement = false;
     }
 
-    @Override
-    public String lineFeed() {
-        return _lineFeed;
-    }
-
     // @since 2.12
     public void writeLeafXsiNilElement(XMLStreamWriter2 sw,
             String nsURI, String localName)
@@ -495,7 +485,7 @@ public class DefaultXmlPrettyPrinter
     /**
      * Dummy implementation that adds no indentation whatsoever
      */
-    protected static class NopIndenter
+    protected class NopIndenter
         implements Indenter, java.io.Serializable
     {
         private static final long serialVersionUID = 1L;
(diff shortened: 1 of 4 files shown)


src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
```diff
@@ -126,6 +133,30 @@ public void indentObjectsWith(Indenter i)
...
+    /**
+     * Sets custom linefeed
+     * @since 2.15
+     */
+    public DefaultXmlPrettyPrinter withCustomlineFeed(String lineFeed) {
+        // 06-Feb-2023, joohyukkim: when JacksonException extends RuntimeExceptions, throw it?
+        if (lineFeed != null) {
+            _lineFeed = lineFeed;
+        }
+        return this;
```
pjfanning, line 157: a non-static method updating a static field - I think this is not good practice

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
```diff
@@ -126,6 +133,30 @@ public void indentObjectsWith(Indenter i)
...
+            lf = System.getProperty("line.separator");
+        } catch (Throwable t) { } // access exception?
+        _lineFeed = lf;
+    }
+
+    /**
+     * Sets custom linefeed
+     * @since 2.15
+     */
+    public DefaultXmlPrettyPrinter withCustomlineFeed(String lineFeed) {
```
pjfanning, line 152: withCustomLineFeed

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
pjfanning: I would suggest that you should have a static DEFAULT_LINE_FEED variable set in the static initializer. (non-sttaic) _lineFeed is initialised equal to DEFAULT_LINE_FEED but `withCustomLineFeed` can modify it.

JooHyukKim: Makes sense. I will get on it.
Thanks!

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
```diff
@@ -79,10 +79,24 @@
 
     /**
      * By default, will try to set as System.getProperty("line.separator")
```
pjfanning, line 81: Can you add a '.' at the end of each sentence here?

JooHyukKim: @pjfanning Got it

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
```diff
@@ -79,10 +79,24 @@
...
+        } catch (Throwable t) { } // access exception?
+
+        if (lf != null) {
+            DEFAULT_LINE_FEED = lf;
+        } else {
+            DEFAULT_LINE_FEED = "\n"; // incase system has changed property name? line.separator
+        }
+    }
+    
+    protected static String _lineFeed = DEFAULT_LINE_FEED;
```
pjfanning, line 99: `protected String _lineFeed = DEFAULT_LINE_FEED;` -- should not be static

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
JooHyukKim: @pjfanning I have a question.
Even if we make `_lineFeed` field non-static, we have static line-feeding Indenters like `Lf2SpacesIndenter`.
They can't access non-static field of outer class. 🥲

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
JooHyukKim: @pjfanning I have a question. 
Even if we make `_lineFeed` field non-static, we have static line-feeding Indenters like `Lf2SpacesIndenter`. Linefeed happens inside `Lf2SpacesIndenter.writeIndentation()`.
So Indenters can't access non-static field of outer class. 🥲

How can we solve this?

JooHyukKim: My thoughts, line-feed outside of Lf2SpacesIndenter?

pjfanning: Lf2SpacesIndenter could have a constructor that takes an instance of DefaultXmlPrettyPrinter in its constructor. Or we could make Lf2SpacesIndenter non-static.

pjfanning:
I have to ask why you are not ensuring the code compiles? 

The `@Override` should not have been added to the new lineFeed() method. In fact, it should not even be necessary. It is likely that  Lf2SpacesIndenter can access the defaultXmlPrettyPrinter._lineFeed.

JooHyukKim:
> I have to ask why you are not ensuring the code compiles?

- My mistake, won't happen again 👍🏻👍🏻

> The `@Override` should not have been added to the new lineFeed() method. In fact, it should not even be necessary. It is likely that Lf2SpacesIndenter can access the defaultXmlPrettyPrinter._lineFeed.

- I making `Lf2SpacesIndenter` non-static at the moment. Sorry for the confusion

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/XmlPrettyPrinterTest.java
```diff
@@ -188,4 +189,76 @@ public void testMultiLevel172() throws Exception
...
+
+        String xml = _xmlMapper.writer()
+            .with(new DefaultXmlPrettyPrinter().withCustomLineFeed("\n\rLF\n\r"))
+            .with(ToXmlGenerator.Feature.WRITE_XML_DECLARATION)
+            .writeValueAsString(root);
+        // unify possible apostrophes to quotes
+        xml = a2q(xml);
+
+        // with indentation, should get linefeeds in prolog/epilog too
+        assertEquals("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
```
JooHyukKim, line 205: I am not sure if this line should also end with customLineFeed?

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/XmlPrettyPrinterTest.java
```diff
@@ -188,4 +189,76 @@ public void testMultiLevel172() throws Exception
...
+                + "<Company>\n\rLF\n\r"
+                + "  <e>\n\rLF\n\r"
+                + "    <employee>\n\rLF\n\r"
+                + "      <id>abc</id>\n\rLF\n\r"
+                + "      <type>FULL_TIME</type>\n\rLF\n\r"
+                + "    </employee>\n\rLF\n\r"
+                + "  </e>\n\rLF\n\r"
+                + "</Company>\n",
```
JooHyukKim, line 213: I am not sure if this line should also end with customLineFeed also?

pjfanning: yes - all the line feeds should be the same

actually, could we rename the method as `withCustomNewLine`? - I don't think 'line feed' is a good name for this property at all - 'new line' is better

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/XmlPrettyPrinterTest.java
pjfanning: see the answer to your follow up question

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/XmlPrettyPrinterTest.java
JooHyukKim: > yes - all the line feeds should be the same

noted will make sure tests are applied there?

> actually, could we rename the method as withCustomNewLine? - I don't think 'line feed' is a good name for this property at all - 'new line' is better

Agreed and I will change names as below!🙏🏼
- `_lineFeed` -> `_newLine`
- `DEFAULT_LINE_FEED` -> `DEFAULT_NEW_LINE`
- `withCustomLineFeed` -> `withCustomNewLine`

JooHyukKim: @pjfanning So here is an update.

> actually, could we rename the method as withCustomNewLine? - I don't think 'line feed' is a good name for this property at all - 'new line' is better

Done.
All names are changed from `line feed` to `new line`

> yes - all the line feeds should be the same

 May I get your opinion here? It seems like writing custom `new lines` before and after root element might be out of this PR's scope. The 'new lines' before and after the root element are written by method `writePrologLinefeed()` of another instance of `DefaultXmlPrettyPrinter`.

 This instance invoking `writePrologLinefeed()` instance seems to be initialized and used inside ToXmlGenerator which we cannot hook into from neither `XmlMapper().setDefaultPrettyPrinter()` nor `ObjectWriter.with(PrettyPrinter)`.

For references,
- Similar issues I found doing research. #554 
-  Check 👉🏼 [the code here writes BEFORE root element](https://github.com/FasterXML/jackson-dataformat-xml/blob/0e59be67bd3e9cfcd73b2b62a95bcc0f5b2dda3c/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java#L258) and [the code here writes AFTER root element](https://github.com/FasterXML/jackson-dataformat-xml/blob/0e59be67bd3e9cfcd73b2b62a95bcc0f5b2dda3c/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java#L619)

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
```diff
@@ -126,6 +142,18 @@ public void indentObjectsWith(Indenter i)
 
     public void spacesInObjectEntries(boolean b) { _spacesInObjectEntries = b; }
 
+    /**
+     * Sets custom linefeed
+     * @since 2.15
+     */
+    public DefaultXmlPrettyPrinter withCustomNewLine(String newLine) {
+        // 06-Feb-2023, joohyukkim: when JacksonException extends RuntimeExceptions, throw it?
```
pjfanning, line 150: can you remove this comment?

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/XmlPrettyPrinterTest.java
pjfanning: It makes little sense to me not to use the same new lines throughout the XML doc. I'm now -1 on this PR.

pjfanning (changes requested):

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/XmlPrettyPrinterTest.java
JooHyukKim: > It makes little sense to me not to use the same new lines throughout the XML doc. I'm now -1 on this PR.

Agreed and -1 from me, too. 
Having a different new line configuration it makes little sense to me also.
I will try to check if a reasonable solution exists before actually closing this PR.
Thank you for your review, always! 🙏🏼🙏🏼

cowtowncoder: Is the discrepancy between prolog/epilog and regular content due to Stax API not exposing write methods?
If I remember correctly there was something odd about that handling; maybe I can help here.

cowtowncoder: Oh no actually looks like there is `XmlPrettyPrinter` method `writePrologLinefeed()` which handles this. As well as possible related `writeRootValueSeparator()`.

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
```diff
@@ -77,6 +77,22 @@
      */
     protected boolean _spacesInObjectEntries = true;
 
+    /**
+     * By default, will try to set as System.getProperty("line.separator").
+     * Can later set custom lineFeed with withCustomlineFeed method.
```
cowtowncoder, line 82: `withCustomNewLine()` method...

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
```diff
@@ -77,6 +77,22 @@
...
+     * @since 2.15
+     */
+    private final static String SYSTEM_DEFAULT_NEW_LINE;
+    static {
+        String lf = null;
+        try {
+            lf = System.getProperty("line.separator");
+        } catch (Throwable t) { } // access exception?
```
cowtowncoder, line 90: Let's just catch `Exception`, not `Throwable` (ok unlike we'd get OOME or such but as a general good practice)

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
```diff
@@ -457,7 +485,7 @@ public void writePrologLinefeed(XMLStreamWriter2 sw) throws XMLStreamException
     /**
      * Dummy implementation that adds no indentation whatsoever
      */
-    protected static class NopIndenter
+    protected class NopIndenter
```
cowtowncoder, line 488: No need to make non-static, let's leave `static`...

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
```diff
@@ -473,7 +501,7 @@ public NopIndenter() { }
      * single space for indentation. It is used as the default
      * indenter for array values.
      */
-    protected static class FixedSpaceIndenter
+    protected class FixedSpaceIndenter
```
cowtowncoder, line 504: Same here, keep `static` (no need to refer to enclosing class)

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
```diff
@@ -501,35 +529,25 @@ public void writeIndentation(JsonGenerator g, int level) throws IOException
...
-            try {
-                lf = System.getProperty("line.separator");
-            } catch (Throwable t) { } // access exception?
-            SYSTEM_LINE_SEPARATOR = (lf == null) ? "\n" : lf;
-        }
-
         final static int SPACE_COUNT = 64;
-        final static char[] SPACES = new char[SPACE_COUNT];
-        static {
+        final char[] SPACES = new char[SPACE_COUNT];
```
cowtowncoder, line 538: Why would this be allocated separately for each instance? Let's keep `static`

JooHyukKim:
Sorry it took me long to react! 🙏🏼
I just moved house 🥲🥲🥲 and it's taking so much time to pack and unpack stuff 

I just got internet back, I will try to apply your feed back asap.

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
JooHyukKim: > Or are non-static inner classes not allowed to have static members? (that'd be the reason)

Yeah, java does not allow non-static inner classes having static members.
But can we move `char[] SPACES`  to outer `DefaultXmlPrettyPrinter` class?

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/XmlPrettyPrinterTest.java
JooHyukKim: > Oh no actually looks like there is `XmlPrettyPrinter` method `writePrologLinefeed()` which handles this. As well as possible related `writeRootValueSeparator()`.

@cowtowncoder The BIG (I think) problem here is that the `DefaultXmlPrettyPrinter` instance running `writePrologLinefeed()` is not same as the one set by `_xmlMapper..with(new DefaultXmlPrettyPrinter().withCustomNewLine("\n\rLF\n\r"))`

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
cowtowncoder: Yes, moving it there probably makes sense to avoid allocating these instances.

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/XmlPrettyPrinterTest.java
cowtowncoder: @JooHyukKim Ok that sounds like a problem indeed! I wonder if it's the case of `Indenter` instance not being created yet (with `createInstance()`) or what is happening.

JooHyukKim: @cowtowncoder 
Not the indenter, actually. There is a separate instance of `DefaultXmlPrettyPrinter` initialized together with `ToXmlGenerator` that invokes `_xmlPrettyPrinter.writePrologLinefeed` to write prolog and epilog.

You can refer to below to code lines.
- [ToXmlGenerator. initGenerator](https://github.com/FasterXML/jackson-dataformat-xml/blob/0e59be67bd3e9cfcd73b2b62a95bcc0f5b2dda3c/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java#L258) method for prolog and
- [ToXmlGenerator._handleObject](https://github.com/FasterXML/jackson-dataformat-xml/blob/0e59be67bd3e9cfcd73b2b62a95bcc0f5b2dda3c/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java#L619) for epilog

My first (probable) observation is that writing prolog and epilog with instance of `DefaultXmlPrettyPrinter` I mentioned above seems NOT to be overridden by `ObjectMapper.setDefaultPrettyPrinter(PrettyPrinter)` nor `ObjectWriter.with(PrettyPrinter)`.

Second observation is `ToXmlGenerator` writes prolog during the process of initialization within `ToXmlGenerator.initGenerator` method.

It will be great, @cowtowncoder if you can help out a little here? To override DefaultXmlPrettyPrinter of `ToXmlGenerator._xmlPrettyPrinter` before it writes prolog and epilog?

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
JooHyukKim: Done :)

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
JooHyukKim: Done~

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
JooHyukKim: > Yes, moving it there probably makes sense to avoid allocating these instances.

DONE! Moved `final static char[] SPACES` to DefaultXmlPrettyPrinter.

JooHyukKim:
All reviews made up to this point are applied to the changes.

The only thing left is how to, or whether to apply custom new line while writing XML declaration with `ToXmlGenerator.Feature.WRITE_XML_DECLARATION` or not.

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/XmlPrettyPrinterTest.java
cowtowncoder: Right, pretty printer instance. I will try to see if I can figure it out -- the problem is that XML module extends basic `PrettyPrinter` API to its needs but databind still tries to orchestrate handling. So I will try to see if I can figure out how to avoid there being different instance being used.

cowtowncoder: @JooHyukKim Ah. Part of the problem is that you haven't updated copy constructor:

    protected DefaultXmlPrettyPrinter(DefaultXmlPrettyPrinter base) {  }

to copy `_newLine` property -- so when method 

```
    @Override
    public DefaultXmlPrettyPrinter createInstance() {
```

is called the new instance "forgets" `_newLine`. So you need to add 

```
    _newLine = base._newLine;
```

in that constructor.

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/XmlPrettyPrinterTest.java
```diff
@@ -188,4 +189,75 @@ public void testMultiLevel172() throws Exception
...
+
+        String xml = _xmlMapper.writer()
+            .with(new DefaultXmlPrettyPrinter())
+            .with(ToXmlGenerator.Feature.WRITE_XML_DECLARATION)
+            .writeValueAsString(root);
+        // unify possible apostrophes to quotes
+        xml = a2q(xml);
+
+        // with indentation, should get newLines in prolog/epilog too
+        assertEquals("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
```
cowtowncoder, line 229: Need to replace `\n` with system linefeed so test passes on Windows systems too.

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
```diff
@@ -126,6 +147,17 @@ public void indentObjectsWith(Indenter i)
 
     public void spacesInObjectEntries(boolean b) { _spacesInObjectEntries = b; }
 
+    /**
+     * Sets custom new-line
+     * @since 2.15
+     */
+    public DefaultXmlPrettyPrinter withCustomNewLine(String newLine) {
+        if (newLine != null) {
+            _newLine = newLine;
```
cowtowncoder, line 156: Why not set `_newLine` to `SYSTEM_DEFAULT_NEW_LINE` if `null` is passed? This would allow reverting back to system newline.

JooHyukKim: Agreed, I don't see any problem doing so.
done!

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/XmlPrettyPrinterTest.java
JooHyukKim: Right, done! 🙏🏼

And plus I checked other similar test cases' assertions and some hard coded `\n` as system linefeed. 
@cowtowncoder if you agree, I will make a separate PR on "replacing all hard coded `\n` to the value read from ` System.getProperty("line.separator");`"?

src/main/java/com/fasterxml/jackson/dataformat/xml/util/DefaultXmlPrettyPrinter.java
JooHyukKim: late reply, but done 👍🏻

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/XmlPrettyPrinterTest.java
JooHyukKim: > is called the new instance "forgets" `_newLine`. So you need to add

Woooooow thank you for helping out 🥹 @cowtowncoder 
How did I miss that!
Done 🙏🏼

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/XmlPrettyPrinterTest.java
cowtowncoder: Yeah sounds good wrt hard-coded linefeeds

cowtowncoder (approved):

cowtowncoder:
Thank you again @JooHyukKim ! Merged for inclusion in 2.15.0

PR #576, opened 2023-03-05
[DRAFT] deep nesting check

pjfanning:
relies on https://github.com/FasterXML/jackson-core/pull/943

It may be better to doc that StreamReadConstraints is ignored for jackson-dataformat-xml and that users are better off using WstxInputFactory and creating XmlMapper using an instance of WstxInputFactory.

WstxInputFactory.getConfig() gives you a ReaderConfig which already has a default max element depth of 1000.

https://fasterxml.github.io/woodstox/javadoc/6.0/com/ctc/wstx/api/ReaderConfig.html

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/deser/FromXmlParser.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/deser/FromXmlParser.java
@@ -686,9 +686,11 @@ public class FromXmlParser
             switch (t) {
             case START_OBJECT:
                 _parsingContext = _parsingContext.createChildObjectContext(-1, -1);
+                _streamReadConstraints.validateNestingDepth(_parsingContext.getNestingDepth());
                 break;
             case START_ARRAY:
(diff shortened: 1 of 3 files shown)


pjfanning:
closing as you can use woodstox to do the nesting check

cowtowncoder:
Yeah this is tricky. On one hand I agree, use of native facilities is generally better way. At the same time, shared configurability and checking has its benefits; including but not limited to Jackson-specific exceptions, context.

To further complicate things, theoretically nesting levels between actual XML and logical "converted" JSON events can have slight discrepancy too.

So for now I think let's use Jackson's new checks and not change Woodstox defaults. We can revisit this in future as necessary.

pjfanning:
Maybe worth having the 2 ways to check for depth (woodstox and jackson-specific). I'm still going to treat this as a lower priority than adding nesting check for Smile and CBOR formats. Possibly, other dataformats too.

cowtowncoder:
Ah. Yes... String length checks would be much more efficient to apply at Woodstox level.

Issue #584, opened 2023-03-17
Deserialization of `null` String values in Arrays / `Collection`s not working as expected


mbladel:
**Environment: Jackson Dataformat XML 2.14.2**

We are trying to serialize and deserialize arrays or collections of Strings containing both empty and `null` values.
We noticed that `null`s are handled correctly when serializing:
* null values are rendered as `<item />`, while empty strings as `<item></item>`
* enabling `ToXmlGenerator.Feature.WRITE_NULLS_AS_XSI_NIL` works as expected

When parsing this values back `null` elements have become empty (`""`) values.
We tried:
* explicitly enabling `FromXmlParser.Feature.PROCESS_XSI_NIL`, even though it should be enabled by default
* enabling `FromXmlParser.Feature.EMPTY_ELEMENT_AS_NULL`
But we still got empty String values as a result.

Testing showed that with simple String attributes inside a class both null and empty values are correctly handled during deserialization, as expected (see also https://github.com/FasterXML/jackson-dataformat-xml/issues/354).

The behavior of String values inside collections should be aligned to how String attributes inside objects are handled.

Here's an example:
```java
XmlMapper xmlMapper = new XmlMapper();
xmlMapper.enable( ToXmlGenerator.Feature.WRITE_NULLS_AS_XSI_NIL );
xmlMapper.enable( FromXmlParser.Feature.PROCESS_XSI_NIL );

String string = xmlMapper.writeValueAsString( new String[] { "", "test", null, "test2" } );
// string: <Strings><item></item><item>test</item><item xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:nil="true"/><item>test2</item></Strings>
String[] parsed =  xmlMapper.readValue( string, String[].class );
// parsed: ["", "test", "", "test2"]
```

cowtowncoder:
Quick note(s):

`<item />` and `<item></item>` are semantically equivalent in XML and although (some) parsers can indicate difference (Woodstox does) -- and most can produce one or the other explicitly -- it's not a good idea to try to use difference for semantics, due to interoperability concerns.

Having said that, yes, handling of `String` values in Arrays or typed `Collection`s should work the same way as individual properties. Handling of difference between `null` and empty String are tricky, and I think Array/Collection deserializers probably have some shortcuts that would need to be disabled for XML.

mbladel:
Thanks for the feedback.

> `<item />` and `<item></item>` are semantically equivalent in XML

[...]

PR #585, opened 2023-03-17
Fix handling of null values in string Arrays and Collections

mbladel:
Possible fix for https://github.com/FasterXML/jackson-dataformat-xml/issues/584.

As a note, empty and `null` string values in `Map`s were correctly handled even before the change (although requiring the map itself to be wrapped in a POJO). I included a test anyways.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/deser/FromXmlParser.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/deser/FromXmlParser.java
@@ -932,12 +932,11 @@ XmlTokenStream.XML_END_ELEMENT, XmlTokenStream.XML_START_ELEMENT, token));
         switch (token) {
         case XmlTokenStream.XML_END_ELEMENT:
             if (_mayBeLeaf) {
-                // NOTE: this is different from nextToken() -- produce "", NOT null
                 _mayBeLeaf = false;
-                _currToken = JsonToken.VALUE_STRING;
+                _currToken = JsonToken.VALUE_NULL;
                 // 13-May-2020, tatu: [dataformat-xml#397]: advance `index`
                 _parsingContext.valueStarted();
-                return (_currText = "");
+                return (_currText = null);
             }
             _currToken = _parsingContext.inArray() ? JsonToken.END_ARRAY : JsonToken.END_OBJECT;
             _parsingContext = _parsingContext.getParent();
--- a/src/test/java/com/fasterxml/jackson/dataformat/xml/lists/StringListRoundtripTest.java
+++ b/src/test/java/com/fasterxml/jackson/dataformat/xml/lists/StringListRoundtripTest.java
@@ -1,0 +1,170 @@
+package com.fasterxml.jackson.dataformat.xml.lists;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.junit.Test;
+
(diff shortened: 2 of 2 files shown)


src/test/java/com/fasterxml/jackson/dataformat/xml/lists/StringListRoundtripTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+import static com.fasterxml.jackson.dataformat.xml.deser.FromXmlParser.Feature.PROCESS_XSI_NIL;
+import static com.fasterxml.jackson.dataformat.xml.ser.ToXmlGenerator.Feature.WRITE_NULLS_AS_XSI_NIL;
+import static java.util.Arrays.asList;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+// [dataformat-xml#584]
+public class StringListRoundtripTest {
```
cowtowncoder, line 20: Looks good, but one big(ger) ask: in addition to root values (which are actually harder to support, bit less common to use), would it be possible to also add "as-property" cases where String arrays, Collections, Maps are properties of POJOs? That would we can double-check that empty Strings do not accidentally become `null`s, which is my main concern.

src/main/java/com/fasterxml/jackson/dataformat/xml/deser/FromXmlParser.java
```diff
@@ -932,12 +932,11 @@ public String nextTextValue() throws IOException
         switch (token) {
         case XmlTokenStream.XML_END_ELEMENT:
             if (_mayBeLeaf) {
-                // NOTE: this is different from nextToken() -- produce "", NOT null
```
cowtowncoder, line 935: Ok, this explicit comment makes me nervous: it specifically spells out that when asking from `nextTextValue()`, empty element should not be coerced into `null`.

But since all existing unit tests, and new ones too, pass, this may make sense.

One related note: the reason `Map`s work ok is that there is no specialized access to "String-valued maps" (but there is for `String[]` and `Collection<String>`!) to use `JsonParser.nextTextValue()`.

One way to alleviate my concerns is to make sure tests are improved -- I noticed additions which are great.
The main addition I'd like to see is addition of checking of `String[]` / `List<String>` valued POJO properties, with null, "". If those retain values as expected I am comfortable with this change.

src/main/java/com/fasterxml/jackson/dataformat/xml/deser/FromXmlParser.java
```diff
@@ -932,12 +932,11 @@ public String nextTextValue() throws IOException
...
                 _mayBeLeaf = false;
-                _currToken = JsonToken.VALUE_STRING;
+                _currToken = JsonToken.VALUE_NULL;
```
cowtowncoder, line 936: Could you also add one-line comment referencing

    / / [dataformat-xml#584]: retain null values

cowtowncoder: Actually I can do that.

cowtowncoder:
Since I am pushing 2.15.0-rc1 soon, I will actually merge this & add comment. We shoyld be able to get some validation on safety with rc1, more so than waiting until rc2 or 2.15.0 final.

Additional tests would be great but can be another PR.

PR #586, opened 2023-03-19
Additional tests for string arrays and collections

mbladel:
Following up on https://github.com/FasterXML/jackson-dataformat-xml/pull/585 with additional tests as suggested. I've added tests for POJOs with a string array or with a list, both are behaving as expected. I also added a test for a simple POJO with just plain string properties to be extra sure nothing breaks.

Regarding [this comment](https://github.com/FasterXML/jackson-dataformat-xml/pull/585#discussion_r1141076795):
> Ok, this explicit comment makes me nervous: it specifically spells out that when asking from nextTextValue(), empty element should not be coerced into null

I was unsure about that comment too, but it looks like that particular token switch case `XmlTokenStream.XML_END_ELEMENT` only occurs for `null` string values (when `xsi:nil` writing and processing is enabled). The `XmlTokenStream.XML_TEXT` case is the one that gets triggered in all other cases as well as for empty strings.

### DIFF
--- a/src/test/java/com/fasterxml/jackson/dataformat/xml/lists/StringListRoundtripTest.java
+++ b/src/test/java/com/fasterxml/jackson/dataformat/xml/lists/StringListRoundtripTest.java
@@ -20,6 +20,8 @@ import static org.junit.Assert.assertNull;
 public class StringListRoundtripTest {
 	private final static XmlMapper MAPPER = new XmlMapper();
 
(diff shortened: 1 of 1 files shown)


cowtowncoder:
@mbladel First of all, big thank you for these tests!

And yes, wrt null induction, I was wondering about the exact same thing. Conversion of `xsi:nil` is not well commented in code but I think your analysis is spot on.
This assessment makes me much more confident in the solution.

cowtowncoder (approved):

Issue #329, opened 2019-02-05
Jackson ignores JacksonXmlElementWrapper on Stream


ndionisi:
jackson-dataformat-xml version: 2.9.8

When using an `XMLMapper` with `Jdk8Module` to serialize `Stream`s, the serializer seems to ignore annotation `@JacksonXmlElementWrapper` placed on a `Stream`.
For example, if I try to serialize this DTO as XML:
```java
public class StreamBasedDto {
        private Stream<String> data;

        @JacksonXmlElementWrapper(localName = "elements")
        @JacksonXmlProperty(localName = "element")
        public Stream<String> getData() {
            return data;
        }

        public void setData(Stream<String> data) {
            this.data = data;
        }
    }
```
I get the following result:
```xml
<StreamBasedDto>
    <element>foo</element>
    <element>bar</element>
</StreamBasedDto>
```
whereas I was expecting the following:
```xml
<StreamBasedDto>
    <elements>
        <element>foo</element>
        <element>bar</element>
    </elements>
</StreamBasedDto>
```
You can find a Java project reproducing the issue here: https://github.com/ndionisi/jackson-xml-element-wrapper-stream. The project tests that `@JacksonXmlElementWrapper` is properly taken into account when serializing `List`, but not when serializing `Stream`.

cowtowncoder:
I think the problem is that special handling via `XmlBeanSerializerModifier` checks for `TypeUtil.isIndexedType()`, and as things are, `Stream` is not considered container type. Same would be true for things like `Iterator` too.

In theory this could be improved, but one nasty problem wrt Jackson 2.x is that there can not be direct dependency to Java 8 types. Java 8 module could provide metadata, but only thing it could do would be to add `TypeModifier` to make `Stream` `JavaType` be a container type... which might have side effects.

With Jackson 3.0 this could be resolved, however.

cowtowncoder:
Note: similar to #302 (possibly can fix both with one patch)

cowtowncoder:
Realized that support probably requires upgrade of JDK baseline for XML module to be Java 8 -- something that is probably reasonable for 2.11, but can not be done in a patch for 2.10.

cowtowncoder:
Was fixed in 2.16.0 but issue not closed; closing now.

Issue #302, opened 2018-07-20
Unable to serialize top-level Java8 Stream


sunruh:
A `Stream` nested in another object is serialized using a value wrapper just as a collection or array, serializing a `Stream` directly fails when more than one element is available as it tries to write a second root. If the Stream emits only a single element serialization works but without wrapping:
```java
public class StreamTest {
    private static final ObjectMapper OBJECT_MAPPER = new XmlMapper()
            .registerModule(new Jdk8Module());
    
    private static class StreamContainer {
        @JsonProperty
        private Stream<String> stream;
        public StreamContainer(Stream<String> stream) { this.stream = stream; }
    }

    @Test
    public void testTopLevelOneElement() throws JsonProcessingException {
        assertNotEquals("<Head>a</Head>",
                OBJECT_MAPPER.writeValueAsString(Stream.of("a")));
    }
    
    @Test
    public void testTopLevelTwoElements() throws JsonProcessingException {
        try {
            assertEquals("<Head><item>a</item></Head>",
                    OBJECT_MAPPER.writeValueAsString(Stream.of("a", "b")));
        } catch (JsonMappingException e) {
            fail("Trying to output second root?", e);
        }
    }
    
    @Test
    public void testNestedOneElement() throws JsonProcessingException {
        assertEquals("<StreamContainer><stream>a</stream></StreamContainer>",
                OBJECT_MAPPER.writeValueAsString(new StreamContainer(Stream.of("a"))));
    }
    
    
    @Test
    public void testNestedTwoElements() throws JsonProcessingException {
        assertEquals("<StreamContainer><stream>a</stream><stream>b</stream></StreamContainer>",
                OBJECT_MAPPER.writeValueAsString(new StreamContainer(Stream.of("a", "b"))));
    }
}
```

The root element name is `Head` for Stream serialization as it's determined from the actual class name `ReferencePipeline$Head`.

The cause lies in `TypeUtil.isIndexedType(Class<?>)` which just checks for Arrays and Collections and only if this returns true, a field name is written (and _nextName set).

It works for the nested Stream as `_nextName` is set due to the Stream being written as a field (multiple times).

cowtowncoder:
Hmmh. Yes, so the work-around(s) used for `Collection`s will not apply to `Stream`s.
I suspect same problem would apply to `Iterator`s / `Iterable`s as well... although in most cases objects probably implement `Collection`.

[...]

Issue #148, opened 2015-04-16
`@JacksonXmlElementWrapper` not respected when serializing `Iterator`s / `Iterable`s


stanislawosinski:
I'm using Jackson 2.5.2 to serialize objects both to JSON and XML. My use case involves serializing large dynamically generated collections, hence the need to expose an Iterator rather than a List for serialization. 

Everything works perfectly with one snag: `@JacksonXmlElementWrapper` does not seem to be respected when serializing `Iterator` / `Iterable`. I'd expect the following code:

``` java
public class WrapperTest {
  public static void main(String[] args) throws IOException {
    new XmlMapper().writeValue(System.out, new ToSerialize());
  }

  private static class ToSerialize {
    @JsonProperty("item")
    @JacksonXmlElementWrapper(localName = "list")
    public Iterator<String> items() {
      return new Iterator<String>() {
        int item = 4;

        @Override
        public boolean hasNext() {
          return item > 0;
        }

        @Override
        public String next() {
          item--;
          return Integer.toString(item);
        }
      };
    }
  }
}
```

to produce:

``` xml
<ToSerialize><list><item>3</item><item>2</item><item>1</item><item>0</item></list></ToSerialize>
```

but instead I'm getting:

``` xml
<ToSerialize><item>3</item><item>2</item><item>1</item><item>0</item></ToSerialize>
```

Am I doing anything wrong here?

If I replace the `Iterator` with a `List`, the wrapping element gets correctly added, but switching to in-memory lists is not practical in our use case. I could instead return a dummy list that returns my original iterator in the `iterator()` method (and throws `UnsupportedOperationException` in all other methods), but this seems a rather ugly hack.

cowtowncoder:
I can not reproduce this so it has probably fixed somewhere since 2.5.2.
Marking as fixed in 2.11.0.

cowtowncoder:
As per @JooHyukKim I think this is actually being only fixed in 2.16 now.

PR #597, opened 2023-05-13
Support serialization of `Iterator` and `Stream`

JooHyukKim:
resolves below issues

- #302
- #329
- #148
    - modified related test

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/TypeUtil.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/TypeUtil.java
@@ -1,19 +1,23 @@
 package com.fasterxml.jackson.dataformat.xml.util;
 
-import java.util.Collection;
-
 import com.fasterxml.jackson.databind.JavaType;
 
-public class TypeUtil
-{
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.stream.Stream;
+
+public class TypeUtil {
     /**
      * Helper method used for figuring out if given raw type is a collection ("indexed") type;
      * in which case a wrapper element is typically added.
      */
     public static boolean isIndexedType(JavaType type)
     {
-        if (type.isContainerType()) {
-            Class<?> cls = type.getRawClass();
+        Class<?> cls = type.getRawClass();
+        if (type.isContainerType()
+            || Iterator.class.isAssignableFrom(cls)
+            || Stream.class.isAssignableFrom(cls)
+        ) {
             // One special case; byte[] will be serialized as base64-encoded String, not real array, so:
             // (actually, ditto for char[]; thought to be a String)
             if (cls == byte[].class || cls == char[].class) {
@@ -31,7 +35,8 @@ public class TypeUtil
 
     public static boolean isIndexedType(Class<?> cls)
     {
-        return (cls.isArray() && cls != byte[].class && cls != char[].class)
-                || Collection.class.isAssignableFrom(cls);
+        return  (cls.isArray() && cls != byte[].class && cls != char[].class) 
+            || Collection.class.isAssignableFrom(cls) || Iterator.class.isAssignableFrom(cls)
+            || Stream.class.isAssignableFrom(cls);
     }
 }
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/util/TypeUtil.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/util/TypeUtil.java
@@ -15,10 +15,7 @@ public class TypeUtil
     public static boolean isIndexedType(JavaType type)
     {
         Class<?> cls = type.getRawClass();
-        if (type.isContainerType()
-            || Iterator.class.isAssignableFrom(cls)
-            || Stream.class.isAssignableFrom(cls)
-        ) {
+        if (type.isContainerType() || canHandleLikeAnIterable(cls)) {
             // One special case; byte[] will be serialized as base64-encoded String, not real array, so:
             // (actually, ditto for char[]; thought to be a String)
             if (cls == byte[].class || cls == char[].class) {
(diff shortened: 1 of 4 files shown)


src/test/java/com/fasterxml/jackson/dataformat/xml/ser/Jdk8StreamSerialization302Test.java
```diff
@@ -0,0 +1,151 @@
...
+        assertEquals(
+            "<IteratorWrapper329><elements>" +
+                "<element>a</element>" +
+                "<element>b</element>" +
+                "</elements></IteratorWrapper329>",
+            OBJECT_MAPPER.writeValueAsString(wrapper));
+    }
+    
+    /*
+    /**********************************************************
```
pjfanning, line 129: can you not add a test scope dependency on Jdk8Module ?

src/main/java/com/fasterxml/jackson/dataformat/xml/util/TypeUtil.java
```diff
@@ -12,8 +14,11 @@
      */
     public static boolean isIndexedType(JavaType type)
     {
-        if (type.isContainerType()) {
-            Class<?> cls = type.getRawClass();
+        Class<?> cls = type.getRawClass();
+        if (type.isContainerType()
+            || Iterator.class.isAssignableFrom(cls)
+            || Stream.class.isAssignableFrom(cls)
```
pjfanning, line 20: is this needed? I'd have expected Iterator and Stream to have JavaTypes that are container types.

pjfanning: @cowtowncoder would `isCollectionLikeType()` something that could be considered here? `isContainerType()` does not include Streams and Iterators. Since we have a `JavaType` instance here, it would be nice not to have to hardcode the classes.

pjfanning: Jackson supports other languages like Scala and Kotlin. The JavaType abstraction is a much more extensible way to do this logic than hardcoding random Java classes.

pjfanning: We may have no choice but to do the code as you are doing it. It would be interesting to get @cowtowncoder's input.

pjfanning:
@JooHyukKim should we also add deserialization tests? If you feel this is too much for this PR but suspect that we do need more work on the deserialization, could you log a follow up issue?

src/main/java/com/fasterxml/jackson/dataformat/xml/util/TypeUtil.java
cowtowncoder: I am trying to figure out how I feel about this. I feel Iterators etc are not "Collection"-like; perhaps we could/should consider new `IteratorType` for 2.16? It would have element type; not sure what other commonality there should be.

Trick here (well one of them) is to figure out how Scala, Kotlin (and maybe later JDK modules) would augment information -- otherwise we could just add something in `SimpleType`.

For inspiration we could look into `ReferenceType` (or `CollectionLikeType`).

There could then be something like `isIterableType()` in `JavaType` returning true for Map(Like), Collection(Like), Array and IteratorTypes.

Would this be a way forward?

cowtowncoder:
On this PR specifically, we could consider doing:

1. Hard-coded approach for 2.15 or 2.16 initially
2. Clean up later on if and when `JavaType` adds detection functionality.

src/main/java/com/fasterxml/jackson/dataformat/xml/util/TypeUtil.java
pjfanning: Maybe we could merge this PR as it is and then proceed with the JavaType changes independently? We can always uptake the JavaType changes for this use case later.

pjfanning:
@JooHyukKim could you create a shared method for checking the class for these `Iterator.class.isAssignableFrom(cls)  || Stream.class.isAssignableFrom(cls)` ?

This check appears twice so it would be good to share the check (just in case we need to add more classes to the check).

src/main/java/com/fasterxml/jackson/dataformat/xml/util/TypeUtil.java
pjfanning: I see that @cowtowncoder made this point about the JavaType changes in another comment.

pjfanning:
Not a big deal but I wrote https://github.com/FasterXML/jackson-module-scala/pull/636 to try out this problem with Scala Iterators and there is an issue and this fix won't help (for Scala Iterators). That doesn't mean we shouldn't proceed - it just highlights that adding a JavaType based solution in the future would be useful.

src/main/java/com/fasterxml/jackson/dataformat/xml/util/TypeUtil.java
cowtowncoder: Right, +1 for separate approach.

As to `IteratorType`, could also consider `IterableType` (although it has same possible issue wrt naming).
Perhaps `IterationType` would be a better choice.

Also on `MapLikeType`, `CollectionLikeType` -- these would not be super types; Maps are not iterators (nor Collections; nor MapLike/CollectionLike). Partly since not all Map(like) types implement actual iterator/iterable (in fact, most don't).
Or put another way: in case something implements `Iterator` (etc) but is also `Map` (or Map-like), our `JavaType` would be `Map[Like]Type`. Iterator tends to be more an add-on (or mix-in) type.

src/test/java/com/fasterxml/jackson/dataformat/xml/lists/ListSerializationTest.java
```diff
@@ -110,7 +110,7 @@ public void testStringList() throws IOException
     // [dataformat-xml#148]
     public void testIteratorSerialization() throws Exception
     {
-        assertEquals("<Bean148><item>2</item><item>1</item><item>0</item></Bean148>",
```
cowtowncoder, line 113: How did this happen? Isn't this backwards-incompatible change in serialization?

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/Jdk8StreamSerialization302Test.java
```diff
@@ -0,0 +1,128 @@
...
+        public Iterator<String> getData() {
+            return data;
+        }
+
+        public void setData(Iterator<String> data) {
+            this.data = data;
+        }
+    }
+
+    public void testCollectionSerialization() throws JsonProcessingException {
```
cowtowncoder, line 63: Let's change these to `throws Exception` since `JsonProcessingException` is removed from 3.0 and `JacksonException` (that exists) does not extend `IOException` any longer.
So use of plain `Exception` allows cleaner merging.

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/Jdk8StreamSerialization302Test.java
cowtowncoder: Ideally this should not be done, to keep cross-repo dependencies minimal.

src/test/java/com/fasterxml/jackson/dataformat/xml/lists/ListSerializationTest.java
cowtowncoder: Ok, one actual request: It'd be good to have separate test class for Iterators or "Iterable types".
It'd be quite small class but I think would be cleaner.

cowtowncoder (approved):

cowtowncoder (approved):

PR #599, opened 2023-05-24
Use `IterationType` in `TypeUtil`

JooHyukKim:
Make use of new `IterationType` class added by https://github.com/FasterXML/jackson-databind/issues/3950

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/XmlSerializerProvider.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/XmlSerializerProvider.java
@@ -96,7 +96,7 @@ public class XmlSerializerProvider extends DefaultSerializerProvider
                 rootName = _rootNameLookup.findRootName(cls, _config);
             }
             _initWithRootName(xgen, rootName);
-            asArray = TypeUtil.isIndexedType(cls);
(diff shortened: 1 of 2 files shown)


cowtowncoder (approved):

cowtowncoder:
> From what I understand, this solution works for `Iterator` support in Scala module, but if not LMK 👍🏻👍🏻

It may be the case; for Collections, Maps, Scala has its alternatives, but maybe `Iterator` used as-is? @pjfanning would know.

I'll get this merged later today, looks good!

pjfanning:
there will need to be changes in jackson-module-scala to uptake IterationType and it could break existing cases - so I may need to come back and request changes

jackson-module-scala currently treats Scala Iterators as CollectionLikeType and changing that to IterationType will have effects that need to be dealt with

pjfanning:
@cowtowncoder we don't see to have a findIterationSerializer to match findCollectionLikeSerializer - should we have a specialised method for IterationType.

cowtowncoder:
@pjfanning I have been wondering about that... whether it's worth it or not. Main concern would be backwards-compatibility wrt registrations (if calls that used to go for general methods now went to new method, existing code might fail).

cowtowncoder:
@AnakinPt As per note on #622, the issue is that you are deploying an incompatible set of versions of components.

PR #600, opened 2023-05-24
Apply JsonTypeInfo.Value usages from Jackson 3.0 #3953

JooHyukKim:
Fixes breakage when PR https://github.com/FasterXML/jackson-databind/pull/3953 is merged.


pom.xml
```diff
@@ -41,6 +41,12 @@ alternative support for serializing POJOs as XML and deserializing XML as pojos.
     <!-- Extends Jackson (jackson-mapper); requires Stax API (and implementation on
          deploy time), Stax2 API.
       -->
+      <dependency>
```
cowtowncoder, line 44: Should not need to do this, as long as `jackson-base` version `2.16.0-SNAPSHOT` has dependencies to 2.16.0-SNAPSHOT. You can also re-build snapshot locally for `jackson-databind` (with `./mvnw clean install`) and that version should be used without other workarouns.

PR #616, opened 2023-11-28
Fix mismatch in setNextIsUnwrapped(true/false) in XmlBeanSerializerBase#serializeFieldsFiltered()

vmi:
In serializeFields() of XmlBeanSerializerBase, setNextIsUnwrapped(true) is executed in L.200-203, and setNextIsUnwrapped(false) is executed in L.215-219.
However, in serializeFieldsFiltered(), setNextIsUnwrapped(true) is executed in L.282-285, but there is no corresponding setNextIsUnwrapped(false).

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/XmlBeanSerializerBase.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/XmlBeanSerializerBase.java
@@ -294,6 +294,11 @@ public abstract class XmlBeanSerializerBase extends BeanSerializerBase
                         filter.serializeAsField(bean, xgen, provider, prop);
                     }
                 }
+                // Reset to avoid next value being written as unwrapped,
+                // for example when property is suppressed
+                if (i == textIndex) {
+                    xgen.setNextIsUnwrapped(false);
+                }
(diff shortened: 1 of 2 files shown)


pjfanning:
Could you provide a test case?

vmi:
I have no test cases.
I will make it if necessary, but please give me some time.

cowtowncoder:
@vmi One other process thing: since this is your first (but hopefully not last! :) ) contribution, I think we will need CLA from:

and it's usually easiest to fill, sign, print & scan/photo, email to `cla` at fasterxml dot com.
This only needs to be done once, before the first contribution and is good for all other contributions.
Apologies for this extra step but it is something we need as OSS project used by corporations.

Thank you again for contributing this -- looking forward to merging it!

vmi:
I added unit test, and sent signed CLA.

cowtowncoder (approved):

cowtowncoder:
Excellent! CLA successfully received, test looks good; will merge after CI passes.

cowtowncoder:
Since this seems like a safe fix, I backported it in 2.16.1 as well so will get released sooner (we only just started 2.17 branch).

Issue #618, opened 2023-12-04
`ArrayIndexOutOfBoundsException` thrown for invalid ending XML string when using JDK default Stax XML parser


arthurscchan:
In XmlTokenStream::_collectUntilTag method, there is an infinite while loop to loop through the provided XML string (through _xmlReader) character by character. The loop only exits by return statements when a valid character (`XMLStreamConstants.START_ELEMENT`, `XMLStreamConstants.END_ELEMENT` or `XMLStreamConstants.END_DOCUMENT`) is found. If the provided XML string is invalid without those characters, it will continue to loop through the whole XML String and eventually throw `ArrayIndexOutOfBoundsException` when _xmlReader has no more characters that can be returned by the next() method. Besides, there are also some other methods depends on those `END_ELEMENT` to stop looping out-of-bound. The suggested fix could be simply wrapping the `ArrayIndexOutOfBoundsException` with the `JsonParseException`.

```java
        CharSequence chars = null;
        while (true) {
            switch (_xmlReader.next()) {
            case XMLStreamConstants.START_ELEMENT:
                return (chars == null) ? "" : chars.toString();

            case XMLStreamConstants.END_ELEMENT:
            case XMLStreamConstants.END_DOCUMENT:
                return (chars == null) ? "" : chars.toString();

            // note: SPACE is ignorable (and seldom seen), not to be included
            case XMLStreamConstants.CHARACTERS:
            case XMLStreamConstants.CDATA:
                // 17-Jul-2017, tatu: as per [dataformat-xml#236], need to try to...
                {
                    String str = _getText(_xmlReader);
                    if (chars == null) {
                        chars = str;
                    } else  {
                        if (chars instanceof String) {
                            chars = new StringBuilder(chars);
                        }
                        ((StringBuilder)chars).append(str);
                    }
                }
                break;
            default:
                // any other type (proc instr, comment etc) is just ignored
            }
        }
    }
```
We found this issue by OSS-Fuzz and it is reported in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64655,  https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64667 and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64659.

cowtowncoder:
Just to be pedantic: loop is not character-by-character but token-by-token.

But I agree with the fix wrt checking that loop ends if there are no more tokens available.

cowtowncoder:
[...]

PR #619, opened 2023-12-04
Fix possible IndexOutOfBoundsException for issue #618

arthurscchan:
This PR provides a possible fix for issue #618 by adding a checking criterion to the while loop to ensure no out-of-bounds read even if the input is invalid.


pjfanning:
Would it be possible to add a test case based on the fuzz issue - or similar? I don't have permissions to see the oss fuzz issues on this repo.

pjfanning:
@arthurscchan this code does not compile - see
```
Error:  /home/runner/work/jackson-dataformat-xml/jackson-dataformat-xml/src/main/java/com/fasterxml/jackson/dataformat/xml/deser/XmlTokenStream.java:[589,5] missing return statement
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project jackson-dataformat-xml: Compilation failure
```

src/main/java/com/fasterxml/jackson/dataformat/xml/deser/FromXmlParser.java
```diff
@@ -714,7 +714,13 @@ public JsonToken nextToken() throws IOException
             return t;
         }
 
-        int token = _nextToken();
+        int token = 0;
```
pjfanning, line 717: `int token;` should be enough

arthurscchan: Fixed.

arthurscchan:
I have also fixed additional location where unexpected IndexOutOfBoundsException are thrown.

src/main/java/com/fasterxml/jackson/dataformat/xml/XmlFactory.java
```diff
@@ -665,6 +665,8 @@ protected FromXmlParser _createParser(byte[] data, int offset, int len, IOContex
             }
         } catch (XMLStreamException e) {
             return StaxUtil.throwAsParseException(e, null);
+        } catch (IndexOutOfBoundsException e) {
+            throw new JsonParseException(null, "Invalid input data.", e);
```
cowtowncoder, line 669: Huh? No, this seems something that should not be needed: nothing ought to throw `IndexOutOfBoundsException`...

src/main/java/com/fasterxml/jackson/dataformat/xml/deser/FromXmlParser.java
```diff
@@ -714,7 +714,13 @@ public JsonToken nextToken() throws IOException
             return t;
         }
 
-        int token = _nextToken();
+        int token;
+
+        try {
+            token = _nextToken();
+        } catch (IndexOutOfBoundsException e) {
+            throw new JsonParseException(this, "Invalid xml input", e);
```
cowtowncoder, line 722: No, please remove

src/main/java/com/fasterxml/jackson/dataformat/xml/deser/XmlDeserializationContext.java
```diff
@@ -97,13 +98,19 @@ public Object readRootValue(JsonParser p, JavaType valueType,
...
+                return _unwrapAndDeserialize(p, valueType, deser, valueToUpdate);
+            }
+            if (valueToUpdate == null) {
+                return deser.deserialize(p, this);
+            }
+            return deser.deserialize(p, this, valueToUpdate);
+        } catch (IndexOutOfBoundsException e) {
+            // If value is invalid without end character, the deserialize will
+            //     read pass the array bound and throws IndexOutOfBoundException
+            throw new JsonParseException(p, "Invalid value with missing JsonToken.END_OBJECT.", e);
```
cowtowncoder, line 112: No, no, no. We should not be catching `IndexOutOfBoundsException`s; there are indications of issues somewhere else. I am not a big fan of defensive programming in this style unless there's absolutely no other choice (for 3rd party deps)

cowtowncoder:
Earlier fix looked better to me; I am -1 on catching `IndexOutOfBoundsException`s: nothing within implementation should be throwing those.

One way forward would be to add tests to show the problem first (can put them under `src/test/java/.../failing` to prevent CI from failing). And then apply necessary fix.
This could be different PR for tests only, and then fix that solves failures (and moves them under "real" test class package, not `failing/`).

src/test/java/com/fasterxml/jackson/dataformat/xml/deser/invalid_xml_1
```diff
@@ -0,0 +1,144 @@
+<xx>http://www.w3.org/2002/XMLS#PCDATAnnsctae<va><x><n xmlns:x="http://w%w.w3.org/2001/XMLSchema-instance" x:nil="true">
```
cowtowncoder, line 1: These are Fuzzer-generated I assume?

cowtowncoder:
@arthurscchan Ok, modified things a bit. So:

1. Yes, all 3 failures are observable if removing Woodstox dependency
2. I added back variation of one of fixes you suggested, just for SJSXP case: I think that makes sense
3. Will try to see how to address 3rd case (first test): it's bit trickier as I would want to isolate the work-around further

cowtowncoder:
@arthurscchan Ok, was able to change things to catch all 3 SJSXP-dependant failure cases. So I think this PR could be read to go -- all I need is the CLA I mentioned earlier.

One downside is that I don't know of a way to easily verify this with Github Actions (CI) for the repo: it can be tested locally by temporarily commenting out Woodstox-dependency, but that would then prevent tests over Woodstox. I'm sure there must be a way to do this (optional dependencies) with Maven (probably with profiles) but for now this will have to do.

arthurscchan:
Hi @cowtowncoder, thanks for the suggestion and fixes. Sorry, I left them out yesterday night and only have time to handle them now and I see you already fixed them. Yes, I do agree that the problem happened because the OSS-Fuzz setting has no Woodstox dependencies thus it is using the "default" SJSXP. But I think it also points out that if the user of this Jackson library is not aware of this "requirement", it is possible that those problems that happened to SJSXP do throw up to their applications and become "unexpected" exceptions. But adding support to SJSXP may be digging into a rabbit hole, thus maybe adding some description in the documentation will be a better alternative.

BTW, I have sent the CLA this morning. Please do let me know if more actions or comments are needed from me on the CLA and those fixes PRs. Thanks.

cowtowncoder:
@arthurscchan No problem at all wrt work -- your baseline was good for figuring out problems.

I concur with your assessment wrt SJSXP as well: ideally we would want to support it to degree possible, and try to avoid most drastic problems (like infinite loops!). But there are limits to how hard to go, given existing bugs (fun fact: I provided half a dozen of bug fixes years ago for pretty basic bugs, when project was open source).

So, I think with CLA I can go ahead and merge this fix.

But one question I have wrt OSS-Fuzz is, whether set up should include Woodstox as a dependency, and if so, when to change. I think there might be value running against SJSXP for little bit more, but for longer term I think Woodstox -- dependency that exists in Maven `pom.xml`, so users really should be getting it if using Maven  or Gradle, unless they explicitly block it, is what we want to test against.

Is this change (to oss-fuzz project set up) something you could help with? If source build is needed, it's from here:

https://github.com/FasterXML/woodstox/

and default branch should be fine.

arthurscchan:
@cowtowncoder I could definitely help on the OSS-Fuzz project set up, I will investigate on the changes. Thanks again for your comments and changes.

cowtowncoder:
@arthurscchan Thank you for digging deep and figuring out these Fuzz failures! This is very valuable and increases value of fuzzing a lot. NPEs, for collection datatypes, for example, are great to catch. And of course edge conditions for format codecs too.

cowtowncoder:
On `jackson-dataformat-xml`, I wonder what changed tho -- I don't think it used to default to SJSXP. Or maybe fuzzing set up was changed to do different kinds of operations.

PR #633, opened 2024-01-31
Fix #324: support auto-detection, conversion of `xsi:type` on generation

### LINKED ISSUE #324: Support use of `xsi:type` for polymorphic serialization (`ToXmlGenerator.Feature.AUTO_DETECT_XSI_TYPE`)
Hey,

I have two subclasses like the following:

```
public abstract class Foo{}
public class Foo1 extends Foo {}
public class Foo2 extends Foo {}
```

And I want to generate XML as the following:
```
<Foo xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:type="Foo1"></Foo>
<Foo xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:type="Foo2"></Foo>
```
Notice there is xsi:type field in the XML.

In JAXB, I could do:
`JAXBContext.newInstance(Foo.class, Foo1.class, Foo2.class);`
Which will provide the xsi:type tag.

How would I do this in Jackson?
Many thanks

cowtowncoder:


cowtowncoder:
Works finally... Hooray!

PR #640, opened 2024-02-19
Add EMPTY_ELEMENT_AS_EMPTY_ARRAY Feature

Croway:

PR #642, opened 2024-03-02
Serialize empty tags to actual json array/object

Croway:
Hello @cowtowncoder , I think it would be nice to serialize empty tag into actual array/objects (in some cases), not text value, in particular, as you can see from the added test case `<name><first/><last>[]</last></name>` will be serialized as `{"first":[],"last":"[]"}` when using JsonNode, the empty tag first is serialized as an empty json array, not a text value.

But I feel like that my [solution](https://github.com/FasterXML/jackson-dataformat-xml/compare/2.17...Croway:json-array-serialize?expand=1#diff-654673c980213587a428541c0914c8096fe61324d9852688d7b04853fa277b45R870) is a bit hacky, do you have any hint?

Once https://github.com/FasterXML/jackson-dataformat-xml/pull/640 is merged I'll rebase this one.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/XmlFactory.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/XmlFactory.java
@@ -66,6 +66,9 @@ public class XmlFactory extends JsonFactory
 
     protected String _cfgNameForTextElement;
 
+    /**
+     * @since 2.17
+     */
     protected String _cfgValueForEmptyElement;
 
     protected XmlNameProcessor _nameProcessor;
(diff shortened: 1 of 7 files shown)


cowtowncoder:
Yeah I am not sure I'm inclined to add support for this. Jackson, in general, tries to steer clear of 2 general areas:

1. Validation of logical content (aside from well-formedness and needing to map between data, Java Objects) -- Bean Validation API to be used for validation, constraints-checking
2. Structural transformations

Obviously there's some amount of functionality that overlaps here, but in general complexity of structural transformations is such, for streaming/incremental reading/writing that it seems best left out.
Things that are needed are already too complex.

PR #651, opened 2024-05-10
Fix #509: Do not throw when parsing mixed content arrays

wickstopher:
Fixes #509.

PR #652, opened 2024-05-10
Fix #509 (re-created #651 with changes)

### LINKED ISSUE #509: Exception when parsing `List`s of mixed content (since 2.13.0)
In Jackson 2.12 the following junit example test (written in Kotlin) works as I'd expect it to

```kotlin
@JsonIgnoreProperties(ignoreUnknown = true)
data class Tspan(
    @field:JacksonXmlText
    var content: String? = null
)

@JsonIgnoreProperties(ignoreUnknown = true)
class Text(
    @field:JacksonXmlElementWrapper(useWrapping = false)
    @field:JacksonXmlProperty(localName = "tspan")
    var spans: List<Tspan>? = null
)

class ExampleTest {
    @Test
    fun svgExample() {
        val text = XmlMapper().readValue("<text>Mixed <tspan>content</tspan> should<tspan> work</tspan></text>", Text::class.java)

        assertThat(text.spans, hasSize(2))
    }
}
```

But when I upgrade to 2.13, I get the following error

```
Unexpected non-whitespace text (' should' in Array context: should not occur (or should be handled)
 at [Source: (StringReader); line: 1, column: 49] (through reference chain: com.visualdx.api.kmax.graphql.Text["tspan"])
com.fasterxml.jackson.databind.JsonMappingException: Unexpected non-whitespace text (' should' in Array context: should not occur (or should be handled)
 at [Source: (StringReader); line: 1, column: 49] (through reference chain: com.visualdx.api.kmax.graphql.Text["tspan"])
 ```

This feels like a bug, but I'm not sure if there's some workaround or setting I'm missing somewhere.

cowtowncoder:
Looks like a can't push to PR, re-creating to add release notes.

### DIFF
--- a/release-notes/CREDITS-2.x
+++ b/release-notes/CREDITS-2.x
@@ -243,3 +243,11 @@ Motonori IWAMURO (vmi@github)
 * Contributed fix for #616: Fix mismatch in `setNextIsUnwrapped(boolean)` in
   `XmlBeanSerializerBase#serializeFieldsFiltered()`
  (2.16.1)
+
+ Dennis Cornwell (@cornwe19)
+
+* Reported #509: Exception when parsing `List`s of mixed content (since 2.13.0)
+
+Christopher R. Wicks (@wickstopher)
+
+* Contributed fix for #509: Exception when parsing `List`s of mixed content (since 2.13.0)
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Fixed for inclusion in:

* 2.16.3
* 2.17.2
* 2.18.0

whenever those are released (may be a while as 2.17.2 was just released; 2.16 has only this fix in branch)

PR #659, opened 2024-06-24
Regression test for databind#4595

SimonCockx:


src/main/java/com/fasterxml/jackson/dataformat/xml/JacksonXmlModule.java
```diff
@@ -1,5 +1,6 @@
 package com.fasterxml.jackson.dataformat.xml;
 
+import com.fasterxml.jackson.core.json.PackageVersion;
```
SimonCockx, line 3: I was getting some import errors on clone. I needed to add these.

cowtowncoder:
@SimonCockx Test failed because of trying to register module after using mapper :-)

Fixing.

cowtowncoder:
Ok, so test passes against `2.17.2-SNAPSHOT`, fails against `2.17.1`. So I think that fix in databind works as expected.

cowtowncoder:
@SimonCockx thank you for contributing this: merged with some changes, and ended up writing separate test in jackson-databind for merging as well.

This is a good fix to get since it will affect quite a few other introspection methods as well.

PR #669, opened 2024-09-02
Fix flaky

Haoranli802:
So this testPullRequest616 test is flaky because the JacksonXml and Xml mapper does not guarantee a fixed order of the elements in the map every time it is used. This test assumes the order will be alphabetical, but it does not follow the alphabetical rule every time the test is run. What I change is setting up local names for the item's elements and a sorting rule for the JacksonXml mapper to be alphabetical. Then, every time this test was triggered, it would follow the same rule and would be flaky.


src/test/java/com/fasterxml/jackson/dataformat/xml/ser/TestSerializationWithFilter.java
```diff
@@ -18,11 +21,14 @@
 public class TestSerializationWithFilter extends XmlTestBase
 {
     @JsonFilter("filter")
+    @JsonPropertyOrder(alphabetic=true)
     static class Item
     {
-        @JacksonXmlText
+        @JacksonXmlProperty(localName = "a")
```
cowtowncoder, line 27: This is not same thing...

PR #672, opened 2024-10-09
Add JDK 17 test functionality

### LINKED ISSUE #671: Add mechanism to run JDK 17 (Records) tests optionally, similar to `jackson-databind`
(note: offshoot from #517) 

Currently we cannot add test cases for Record-handling, in Jackson 2.x, due to JDK baseline being Java 8 still.
But there is a way to do this; used f.ex by `jackson-databind`.
This is relatively simple by:

1. Adding tests under `src/test-jdk17/java/`
2. Using specific Maven profile to conditionally include this test source directory if (and only if) running on JDK 17+

It'd be good to be able to add specific Records-with-XML test cases.

JooHyukKim:
resolves #671

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -202,4 +202,58 @@ alternative support for serializing POJOs as XML and deserializing XML as pojos.
     </repository>
   </repositories>
 
+  <profiles>
+    <profile>
+      <!-- And different set up for JDK 17 -->
+      <id>java17</id>
+      <activation>
+        <jdk>17</jdk>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>build-helper-maven-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>add-test-source</id>
+                <phase>generate-test-sources</phase>
+                <goals>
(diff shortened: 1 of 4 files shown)


JooHyukKim:
Tried manually failing jdk 17 to see if it works via https://github.com/FasterXML/jackson-dataformat-xml/pull/672/commits/e60e5f96b8de2009424f69f3c5d5103536ae0bb0. Works fine 👌🏼

cowtowncoder (approved):

cowtowncoder:
Oh shoot. This was agaisnt 2.17 for some reason? I don't think we really needed that... although I guess that's ok.

JooHyukKim:
@cowtowncoder My bad 🙉 I forgot to point out about the base branch.

extension modules tend to get updated less frequently so there will be more people reporting issues from earlier version, so I figured might as well get merged into reasonably earlier version.

cowtowncoder:
My bad for not asking for re-base. Problem/challenge here was just merging, `pom.xml` does not merge cleanly so had to do more work.

We are very unlikely to backport fixes in 2.17 for any modules.

Still, it's all good, got it merged and we have it even for 2.17 :)

PR #680, opened 2024-10-29
Fix #678: allow configuring custom SerializerFactory via JacksonXmlModule

### LINKED ISSUE #678: XML module not registered correctly when setting a custom `SerializerFactory`
I'm using a custom `SerializerFactory` as follows.
```
ObjectMapper mapper = new XmlMapper().setSerializerFactory(MySerializerFactory.INSTANCE);
```

I noticed that in the constructor of `XmlMapper`, the default `XML_MODULE` is registered, which also configures the `SerializerFactory`. Since I'm overriding it after constructing, this configuration is lost, resulting in different behaviour, which is very confusing.

I currently don't see any way for registering my custom `SerializerFactory` *before* the `XML_MODULE` is registered. Is this intended? It seems error-prone.

My current workaround:
```
ObjectMapper mapper = new XmlMapper((JacksonXmlModule) null)
    .setSerializerFactory(RosettaSerialiserFactory.INSTANCE)
    .registerModule(new JacksonXmlModule());
```

cowtowncoder:
Implements a feature to address #678.

### DIFF
--- a/release-notes/VERSION-2.x
+++ b/release-notes/VERSION-2.x
@@ -4,6 +4,11 @@ Project: jackson-dataformat-xml
 === Releases ===
 ------------------------------------------------------------------------
 
+2.18.2 (not yet released)
+
+#678: XML module not registered correctly when setting a custom `SerializerFactory`
+ (reported by @SimonCockx)
+
 2.18.1 (28-Oct-2024)
 
 No changes since 2.18.0
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/JacksonXmlModule.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/JacksonXmlModule.java
@@ -2,7 +2,7 @@ package com.fasterxml.jackson.dataformat.xml;
 
 import com.fasterxml.jackson.databind.AnnotationIntrospector;
 import com.fasterxml.jackson.databind.module.SimpleModule;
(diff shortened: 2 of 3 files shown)


cowtowncoder:
@SimonCockx Ok so I don't think there is (alas!) any way to fix obvious handling to allow direct setting via `ObjectMapper` -- but it seems straight-forward to add functionality to pass `SerializerFactory` as module configuration like so:

```
ObjectMapper mapper = XmlMapper.builder()
    .addModule(new JacksonXmlModule().overrideSerializerFactory(RosettaSerialiserFactory.INSTANCE))
    .build();
```

which should then set `SerializerFactory` before module registration.

This PR is for adding that functionality, although in the meantime your work-around seems legit (if non-obvious).

LMK what do you think. Addition would go in `2.18.2` which is some way off as I just released 2.18.1 earlier today.

cowtowncoder:
I think this makes sense, will merge.

SimonCockx:
@cowtowncoder I completely missed this, sorry. Thanks for the PR, it looks good!

cowtowncoder:
No problem, figured there's not much downside so could merge it. But will take a while to be available -- I did add it in 2.18 (minor deviation from SemVer), but since 2.18.1 was just released 2.18.2 won't be immediate, will be a month or two, likely.

PR #707, opened 2025-01-17
Try using shared test util classes

cowtowncoder:
NOTE: does not yet try using `@JacksonTestFailureExpected` since "failing" not yet converted to "tofix".


cowtowncoder:
@JooHyukKim This is the strange part: use from `jackson-databind` works fine (even if `test-jar` is missing `module-info.class`); but not from this module. I have no idea what is different, or how to work around the problem.. :-(

Issue #700, opened 2025-01-11
JSTEP-10: Unifying testing structure/tools


JooHyukKim:
This issue serves as an umbrella issue for resolving [[JSTEP-10: Unifying testing structure/tools]](https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-10)

cowtowncoder:
Almost there: I think migrating "failing" to "tofix" is the only thing missing here. Right @JooHyukKim ?

PR #709, opened 2025-01-20
[JSTEP-10] Move `/failing` tests

JooHyukKim:


pom.xml
```diff
@@ -189,9 +189,6 @@ alternative support for serializing POJOs as XML and deserializing XML as pojos.
           <artifactId>maven-surefire-plugin</artifactId>
           <version>${version.plugin.surefire}</version>
           <configuration>
-            <excludes>
-              <exclude>**/failing/**/*.java</exclude>
-            </excludes>
             <includes>
               <include>**/Test*.java</include>
               <include>**/*Test.java</include>
```
JooHyukKim, line 194: I'm not sure if we still need these 🤔

cowtowncoder: No I don't think so, `@Test` is better

cowtowncoder (approved):

pom.xml
cowtowncoder: I will remove after merging.

cowtowncoder:
Merged, #700 no complete!

PR #736, opened 2025-03-17
java17: show that test does not run

pjfanning:
just demos that the jdk17 tests don't run


pjfanning:
Closing as #737 seems to fix this

cowtowncoder:
Yeah, JUnit4 tests can't co-exist with JUnit5 ones I think.

PR #737, opened 2025-03-17
fix test imports (jupiter)

pjfanning:
This is a bug fix because tests that use the org.junit.Test annotation no longer seem to run.

See #736


cowtowncoder (approved):
LGTM

PR #738, opened 2025-03-17
run jdk17 tests on all Java versions 17+

pjfanning:
current maven setup means JDK17 tests are not run when we test with Java 21 - only when Java version is exactly 17.


cowtowncoder (approved):

cowtowncoder:
Good catch!

Issue #734, opened 2025-03-13
XmlMapper maps null value of xml element with atribute


stefanCotruta:
### Search before asking

- [x] I searched in the [issues](https://github.com/FasterXML/jackson-databind/issues) and found nothing similar.

### Describe the bug

Jackson 2.17.2 correctly maps the xml element containing an attribute to a Java record. If I use 2.18.2, it only recognizes the attribute but not the value. 

### Version Information

2.18.2

### Reproduction
```java
@SpringBootApplication
public class DemoApplication {

    public static void main(String[] args) throws JsonProcessingException {
        SpringApplication.run(DemoApplication.class, args);

        String xml = "<Amt Ccy=\"Eur\">1</Amt>";
        XmlMapper xmlMapper = new XmlMapper();
        Amount amount = xmlMapper.readValue(xml, Amount.class);
        System.out.println(amount);
    }
    private record Amount(@JacksonXmlProperty String amt, @JacksonXmlProperty(isAttribute = true, localName = "Ccy") String currency) {}
}
```
```xml
// Using 2.17.2 for both dependencies instead of 2.18.2 works correctly without any code changes
        <dependency>
            <groupId>com.fasterxml.jackson.core</groupId>
            <artifactId>jackson-databind</artifactId>
            <version>2.18.2</version>
        </dependency>
        <dependency>
            <groupId>com.fasterxml.jackson.dataformat</groupId>
            <artifactId>jackson-dataformat-xml</artifactId>
            <version>2.18.2</version>
        </dependency>
``` 

### Expected behavior

Amount[amt=1, currency=Eur]

was instead

UnrecognizedPropertyException: Unrecognized field "" [...], not marked as ignorable (2 known properties: "amt", "Ccy"]

cowtowncoder:
+1 to what @pjfanning said.

stefanCotruta:
> Please do not provide screen shots. Provide text. Please also provide the Amount class definition so the issue can be reproduced.

I updated the post, hope this helps.

cowtowncoder:
Thanks. 2 more things:

1. Should not depend on Spring Boot -- but I think parts can be just dropped.
2. Instead of `System.out.println(amount);` would it be possible to either use assertions, or include observed output here?

Also: I assume 2.18.3 is same as 2.18.2 (was recently released, no XML-specific changes).

stefanCotruta:
I updated it. The issue is on 2.18.3 as well.

[...]

PR #739, opened 2025-03-17
add test for issue 734 (disabled)

pjfanning:
@cowtowncoder this seems to reproduce the issue in #734 

I targeted master because of the build issues in 2.19 branch - #737 (etc)

When enabled, I get
```
[ERROR] Errors: 
[ERROR]   XmlRecordDeser734Test.testDeser:25 » InvalidDefinition Invalid definition for property '' (of type `tools.jackson.dataformat.xml.tofix.records.XmlRecordDeser734Test$Amount`): Could not find creator property with name '' (known Creator properties: [value, Ccy])
 at [Source: (String)"<Amt Ccy="EUR">1</Amt>"; line: 1, column: 1]
```


cowtowncoder:
Could perhaps now be backported?

Issue #291, opened 2018-05-19
Problems with Builder pattern with XML


cowtowncoder:
(note: moved from https://github.com/FasterXML/jackson-module-jaxb-annotations/issues/65 filed by @shahmalav )

Below is my POJO using builder pattern for Address. 

```
@JsonDeserialize(builder = Address.AddressBuilder.class)
public final class Address {

    private final String address1;
    private final String city;
    private final String stateProvince;
    private final String postalCode;
    private final Country country;
    private final String county;

    private Address(AddressBuilder addressbuilder) {
        this.address1 = addressbuilder.address1;
        this.city = addressbuilder.city;
        this.stateProvince = addressbuilder.stateProvince;
        this.postalCode = addressbuilder.postalCode;
        this.country = addressbuilder.country;
        this.county = addressbuilder.county;
    }

    public String getAddress1() {
        return address1;
    }
    public String getCity() {
        return city;
    }

    public String getStateProvince() {
        return stateProvince;
    }

    public String getPostalCode() {
        return postalCode;
    }

    public Country getCountry() {
        return country;
    }

    public String getCounty() {
        return county;
    }

    @JsonPOJOBuilder(withPrefix = "set")
    public static class AddressBuilder {
     
        @XmlElement(name = "Address1")
        private final String address1;

        @XmlElement(name = "City")
        private final String city;

        @XmlElement(name = "StateProvince")
        private final String stateProvince;

        @XmlElement(name = "PostalCode")
        private final String postalCode;

        @XmlElement(name = "Country")
        private final Country country;

        @XmlElement(name = "County")
        private String county;
        
        @JsonCreator
        public AddressBuilder(@JsonProperty("address1") String address1, @JsonProperty("city") String city, @JsonProperty("stateProvince") String stateProvince,
                @JsonProperty("postalCode") String postalCode, @JsonProperty("country") Country country) {
            this.address1 = address1;
            this.city = city;
            this.stateProvince = stateProvince;
            this.postalCode = postalCode;
            this.country = country;
        }

        public AddressBuilder setCounty(String county) {
            this.county = county;
            return this;
        }

        public Address build() {
            return new Address(this);
        }
    }
}
```
I can serialize and deserialize above class using JSON but when it comes to XML, it throws an exception.

Following is the stack trace:

```
[...]

PR #744, opened 2025-04-11
Update test for #291

stevenschlansker:
Currently the case is mis-matched which accidentally relies on `final` field setting.
Instead, just fix the annotations so the case matches and the constructor
is properly used


cowtowncoder:
I think use of `final` here may be an oversight: would it be possible to instead remove `final` markers to test regular Builder pattern for XML module?
(unless there are already similar tests)

stevenschlansker:
Ok, does this look better?

cowtowncoder:
@stevenschlansker Yes, thank you! I think I'll backport this in 2.19 as well.

PR #745, opened 2025-04-12
Add feature to include `standalone='yes'` in xml declaration

duoduobingbing:
By default there is no easy way with Jackson XML to include `standalone='yes'` inside the XML declaration. It requires swapping out the `xmlWriter` which is really cumbersome because the implementations of `BaseStreamWriter` from Woodstox are final like `RepairingNsStreamWriter`. Hence to change this the entirety of e.g. `RepairingNsStreamWriter` needs to be copied to the project only to override `BaseStreamWriter#writeStartDocument` to call `doWriteStartDocument(version, encoding, "yes");` instead of the default `doWriteStartDocument(version, encoding, null);`.

This PR adds `ToXmlGenerator.Feature.WRITE_STANDALONE_YES_TO_XML_DECLARATION` which is `false` by default.

When enabled `standalone='yes'` will be added to the XML declaration if including it is enabled as well: 
```java
XmlMapper mapper = new XmlMapper();
mapper.configure(ToXmlGenerator.Feature.WRITE_XML_DECLARATION, true);
mapper.configure(ToXmlGenerator.Feature.WRITE_STANDALONE_YES_TO_XML_DECLARATION, true);
String xml = mapper.writeValueAsString(Map.of("abc","def"));
```
will cause the first line to have<br>
`<?xml version='1.0' encoding='UTF-8' standalone='yes'?>`

There is definetely a need for this as demonstrated by these posts:
* https://stackoverflow.com/questions/74195415/springboot-jackson-2-13-x-outputting-standalone-yes-in-xml-header
* https://stackoverflow.com/questions/52999562/add-standalone-directive-to-jackson-xml-serialization
* https://www.mail-archive.com/jackson-dev@googlegroups.com/msg00345.html

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java
@@ -52,6 +52,15 @@ public class ToXmlGenerator
         WRITE_XML_DECLARATION(false),
 
         /**
+         * Feature that controls whether XML declaration should include the standalone attribute
+         * when generator is initialized (true) or not (false). Only honored when
+         * {@link Feature#WRITE_XML_DECLARATION WRITE_XML_DECLARATION} is enabled
+         *
+         * @since 2.19
+         */
+        WRITE_STANDALONE_YES_TO_XML_DECLARATION(false),
+
+        /**
          * Feature that controls whether output should be done as XML 1.1; if so,
          * certain aspects may differ from default (1.0) processing: for example,
          * XML declaration will be automatically added (regardless of setting
@@ -297,15 +306,24 @@ public class ToXmlGenerator
         _initialized = true;
         try {
             boolean xmlDeclWritten;
-            if (Feature.WRITE_XML_1_1.enabledIn(_formatFeatures)) {
-                _xmlWriter.writeStartDocument("UTF-8", "1.1");
-                xmlDeclWritten = true;
-            } else if (Feature.WRITE_XML_DECLARATION.enabledIn(_formatFeatures)) {
-                _xmlWriter.writeStartDocument("UTF-8", "1.0");
+
+            if (Feature.WRITE_XML_1_1.enabledIn(_formatFeatures) ||
+                    Feature.WRITE_XML_DECLARATION.enabledIn(_formatFeatures)) {
+
+                String xmlVersion = Feature.WRITE_XML_1_1.enabledIn(_formatFeatures) ? "1.1" : "1.0";
+                String encoding = "UTF-8";
+
+                if (Feature.WRITE_STANDALONE_YES_TO_XML_DECLARATION.enabledIn(_formatFeatures)) {
+                    _xmlWriter.writeStartDocument(xmlVersion, encoding, true);
+                } else {
+                    _xmlWriter.writeStartDocument(encoding, xmlVersion);
+                }
+
                 xmlDeclWritten = true;
             } else {
                 xmlDeclWritten = false;
             }
+
             // as per [dataformat-xml#172], try adding indentation
             if (xmlDeclWritten && (_xmlPrettyPrinter != null)) {
                 // ... but only if it is likely to succeed:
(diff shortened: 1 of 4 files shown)


pjfanning:
Please describe why you even need standalone='yes'. That flag is only supposed to used with DTDs - and hardly anyone uses DTDs any more.

src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java
```diff
@@ -51,6 +51,15 @@ public enum Feature implements FormatFeature
          */
         WRITE_XML_DECLARATION(false),
 
+        /**
+         * Feature that controls whether XML declaration should include the standalone attribute
+         * when generator is initialized (true) or not (false). Only honored when
+         * {@link Feature#WRITE_XML_DECLARATION WRITE_XML_DECLARATION} is enabled
+         *
+         * @since 2.XX
```
pjfanning: 2.19 - like the branch name

src/main/java/com/fasterxml/jackson/dataformat/xml/ser/ToXmlGenerator.java
```diff
@@ -297,15 +306,24 @@ public void initGenerator()  throws IOException
...
+
+                if(Feature.WRITE_STANDALONE_YES_TO_XML_DECLARATION.enabledIn(_formatFeatures)) {
+                    _xmlWriter.writeStartDocument(xmlVersion, encoding, true);
+                }else{
+                    _xmlWriter.writeStartDocument(encoding, xmlVersion);
+                }
+
                 xmlDeclWritten = true;
-            } else {
+            }else {
```
pjfanning: undo the format change - add back the space

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/TestXmlDeclaration.java
```diff
@@ -32,5 +32,26 @@ public void testXml11Declaration() throws Exception
...
+
+    @Test
+    public void testXml11DeclarationWithStandalone() throws Exception
+    {
+        XmlMapper mapper = new XmlMapper();
+        mapper.configure(ToXmlGenerator.Feature.WRITE_XML_DECLARATION, true);
+        mapper.configure(ToXmlGenerator.Feature.WRITE_XML_1_1, true);
+        mapper.configure(ToXmlGenerator.Feature.WRITE_STANDALONE_YES_TO_XML_DECLARATION, true);
+        String xml = mapper.writeValueAsString(new StringBean("abcd"));
+        assertEquals(xml, "<?xml version='1.1' encoding='UTF-8' standalone='yes'?><StringBean><text>abcd</text></StringBean>");
```
pjfanning: these params are in the wrong order - expected value goes first

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/TestXmlDeclaration.java
```diff
@@ -32,5 +32,26 @@ public void testXml11Declaration() throws Exception
...
+    }
+
+    @Test
+    public void testXml10DeclarationWithStandalone() throws Exception
+    {
+        XmlMapper mapper = new XmlMapper();
+        mapper.configure(ToXmlGenerator.Feature.WRITE_XML_DECLARATION, true);
+        mapper.configure(ToXmlGenerator.Feature.WRITE_STANDALONE_YES_TO_XML_DECLARATION, true);
+        String xml = mapper.writeValueAsString(new StringBean("abcd"));
+        assertEquals(xml, "<?xml version='1.0' encoding='UTF-8' standalone='yes'?><StringBean><text>abcd</text></StringBean>");
```
pjfanning: wrong param order

duoduobingbing:
> Please describe why you even need standalone='yes'. That flag is only supposed to used with DTDs - and hardly anyone uses DTDs any more.

First of all thank you so much for reviewing this so quickly. 

We work with a very nasty consumer on the other end which refuses to read the XML if standalone=yes is missing.

src/test/java/com/fasterxml/jackson/dataformat/xml/ser/TestXmlDeclaration.java
duoduobingbing: fixed

cowtowncoder (approved):

PR #754, opened 2025-04-27
Naive attempt at fixing #306

### LINKED ISSUE #306: Can not use `@JacksonXmlText` for Creator property (creator parameter)
(Refiling after wrong post in [jackson-module-base #57](https://github.com/FasterXML/jackson-modules-base/issues/57))

See this test case:
```
import static org.hamcrest.CoreMatchers.is;
import static org.junit.Assert.assertThat;

import com.fasterxml.jackson.dataformat.xml.JacksonXmlModule;
import com.fasterxml.jackson.dataformat.xml.XmlMapper;
import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement;
import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlText;
import org.junit.Test;

import java.beans.ConstructorProperties;
import java.io.IOException;
import javax.xml.bind.annotation.XmlAttribute;

public class JacksonXmlTextTest {

    private static final String XML = "<ROOT><CHILD attr=\"attr_value\">text</CHILD></ROOT>";

    @JacksonXmlRootElement(localName = "ROOT")
    public static class Root {
        @JacksonXmlProperty(localName = "CHILD") final Child child;

        public Root(Child child) {
            this.child = child;
        }

        public Child getChild() {
            return child;
        }
    }

    public static class Child {
        @XmlAttribute
        String attr;

        @JacksonXmlText
        String el;

        public Child() {
        }

        @ConstructorProperties({"attr", "el"})
        private Child(String attr, String el) {
            this.attr = attr;
            this.el = el;
        }

        public String getAttr() {
            return this.attr;
        }

        public String getEl() {
            return this.el;
        }

        public void setAttr(String attr) {
            this.attr = attr;
        }

        public void setEl(String el) {
            this.el = el;
        }
    }

    @JacksonXmlRootElement(localName = "ROOT")
    public static class RootWithoutConstructor {
        @JacksonXmlProperty(localName = "CHILD") final ChildWithoutConstructor child;

        @ConstructorProperties({"child"})
        public RootWithoutConstructor(ChildWithoutConstructor child) {
            this.child = child;
        }

        public ChildWithoutConstructor getChild() {
            return child;
        }
    }

    public static class ChildWithoutConstructor {
        @XmlAttribute
        String attr;

        @JacksonXmlText
        String el;

        public String getAttr() {
            return this.attr;
        }

        public String getEl() {
            return this.el;
        }

        public void setAttr(String attr) {
            this.attr = attr;
        }

        public void setEl(String el) {
            this.el = el;
        }
    }

    @Test
    public void verifyJacksonXmlTextIsUnmarshalled() throws IOException {
        final XmlMapper mapper = new XmlMapper(new JacksonXmlModule());
        assertThat(
            mapper.readValue(XML, Root.class).getChild().el,
            is("text")
        );
    }

    @Test
    public void verifyJacksonXmlTextIsUnmarshalledWithoutConstructor() throws IOException {
        final XmlMapper mapper = new XmlMapper(new JacksonXmlModule());
        assertThat(
            mapper.readValue(XML, RootWithoutConstructor.class).getChild().el,
            is("text")
        );
    }

}
```

`verifyJacksonXmlTextIsUnmarshalled()` fails. `verifyJacksonXmlTextIsUnmarshalledWithoutConstructor()` succeeds.

duoduobingbing:
This PR aims to fix #306 naively by just changing the default property name of XmlText from `""` to `<xml:text>`. The new value was chosen, because it is an invalid property and element name in XML and thus should hopefully not collide with existing setups.

I was not sure if XmlText having the property name `""` is considered an implementation detail. If it is not an implementation detail this is most likely to be considered as a breaking change.

Feel free to close this PR if e.g. something else is planned to fix #306 or if XmlText has to keep the property name empty string because of compatibility reasons.

### DIFF
--- a/src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
+++ b/src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
@@ -8,7 +8,6 @@ import tools.jackson.databind.PropertyName;
 import tools.jackson.databind.cfg.MapperConfig;
 import tools.jackson.databind.introspect.*;
 import tools.jackson.dataformat.xml.annotation.*;
-import tools.jackson.dataformat.xml.deser.FromXmlParser;
 
 /**
  * Extension of {@link JacksonAnnotationIntrospector} that is needed to support
@@ -38,12 +37,19 @@ public class JacksonXmlAnnotationIntrospector
 
     protected boolean _cfgDefaultUseWrapper;
 
+    protected final JacksonXmlAnnotationIntrospectorConfig _cfgIntrospectorConfig;
+
     public JacksonXmlAnnotationIntrospector() {
         this(DEFAULT_USE_WRAPPER);
     }
 
     public JacksonXmlAnnotationIntrospector(boolean defaultUseWrapper) {
+        this(defaultUseWrapper, new JacksonXmlAnnotationIntrospectorConfig());
+    }
+
+    public JacksonXmlAnnotationIntrospector(boolean defaultUseWrapper, JacksonXmlAnnotationIntrospectorConfig introspectorConfig) {
         _cfgDefaultUseWrapper = defaultUseWrapper;
+        _cfgIntrospectorConfig = introspectorConfig;
     }
 
     /*
@@ -85,7 +91,7 @@ public class JacksonXmlAnnotationIntrospector
         }
         return null;
     }
-    
+
     @SuppressWarnings("deprecation")
     @Override
     public PropertyName findRootName(MapperConfig<?> config, AnnotatedClass ac)
@@ -94,7 +100,7 @@ public class JacksonXmlAnnotationIntrospector
         if (root != null) {
             String local = root.localName();
             String ns = root.namespace();
-            
+
             if (local.length() == 0 && ns.length() == 0) {
                 return PropertyName.USE_DEFAULT;
             }
@@ -209,8 +215,11 @@ public class JacksonXmlAnnotationIntrospector
         PropertyName pn = PropertyName.merge(_findXmlName(a),
                 super.findNameForDeserialization(config, a));
         if (pn == null) {
-            if(_hasAnnotation(a, JacksonXmlText.class)){
-                return PropertyName.construct(FromXmlParser.DEFAULT_TEXT_PROPERTY);
+            JacksonXmlText jacksonXmlTextAnnotation = _findAnnotation(a, JacksonXmlText.class);
+
+            if (jacksonXmlTextAnnotation != null && jacksonXmlTextAnnotation.value() &&
+                !_cfgIntrospectorConfig.inferXmlTextPropertyName()) {
+                return _cfgIntrospectorConfig.xmlTextPropertyName();
             }
 
             if (_hasOneOf(a, ANNOTATIONS_TO_INFER_XML_PROP)) {
--- a/src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
+++ b/src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
@@ -8,6 +8,7 @@ import tools.jackson.databind.PropertyName;
 import tools.jackson.databind.cfg.MapperConfig;
 import tools.jackson.databind.introspect.*;
 import tools.jackson.dataformat.xml.annotation.*;
+import tools.jackson.dataformat.xml.deser.FromXmlParser;
 
 /**
  * Extension of {@link JacksonAnnotationIntrospector} that is needed to support
@@ -208,6 +209,10 @@ public class JacksonXmlAnnotationIntrospector
         PropertyName pn = PropertyName.merge(_findXmlName(a),
                 super.findNameForDeserialization(config, a));
         if (pn == null) {
+            if(_hasAnnotation(a, JacksonXmlText.class)){
+                return PropertyName.construct(FromXmlParser.DEFAULT_TEXT_PROPERTY);
+            }
+
             if (_hasOneOf(a, ANNOTATIONS_TO_INFER_XML_PROP)) {
                 return PropertyName.USE_DEFAULT;
             }
--- a/src/main/java/tools/jackson/dataformat/xml/deser/FromXmlParser.java
+++ b/src/main/java/tools/jackson/dataformat/xml/deser/FromXmlParser.java
@@ -36,7 +36,7 @@ public class FromXmlParser
 {
     /**
      * The default name placeholder for XML text segments is &lt;xml:text&gt;
-     * @since 3.0.0 Is now &lt;xml:text&gt; - was empty String ("") before.
+     * @since 3.0 Is now &lt;xml:text&gt; - was empty String ("") before.
      */
     public final static String DEFAULT_TEXT_PROPERTY = "<xml:text>";
 
--- a/src/main/java/tools/jackson/dataformat/xml/deser/FromXmlParser.java
+++ b/src/main/java/tools/jackson/dataformat/xml/deser/FromXmlParser.java
@@ -35,8 +35,12 @@ public class FromXmlParser
     implements ElementWrappable
 {
     /**
-     * The default name placeholder for XML text segments is &lt;xml:text&gt;
-     * @since 3.0 Is now &lt;xml:text&gt; - was empty String ("") before.
+     * The default name placeholder for XML text segments: used because Token stream
+     * requires all values inside "Objects" to have names associated.
+     * For Jackson 3.x this is {@code <xml:text>}; in 2.x matching constant was defined
+     * as empty String ({@code ""}).
+     * 
+     * @since 3.0 Constant was renamed: was {@code DEFAULT_UNNAMED_TEXT_PROPERTY} in 2.x
      */
     public final static String DEFAULT_TEXT_PROPERTY = "<xml:text>";
 
--- a/src/main/java/tools/jackson/dataformat/xml/deser/FromXmlParser.java
+++ b/src/main/java/tools/jackson/dataformat/xml/deser/FromXmlParser.java
@@ -35,10 +35,10 @@ public class FromXmlParser
     implements ElementWrappable
 {
     /**
-     * The default name placeholder for XML text segments is empty
-     * String ("").
+     * The default name placeholder for XML text segments is &lt;xml:text&gt;
+     * @since 3.0.0 Is now &lt;xml:text&gt; - was empty String ("") before.
      */
-    public final static String DEFAULT_UNNAMED_TEXT_PROPERTY = "";
+    public final static String DEFAULT_TEXT_PROPERTY = "<xml:text>";
 
     /**
      * XML format has some peculiarities, indicated via capability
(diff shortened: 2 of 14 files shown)


cowtowncoder:
Choice of "" to represent `XmlText` is not a simple implementation detail so I doubt we can change that for Jackson 2.x, unfortunately due to (like you correctly guessed) compatibility concerns.

But I think we might be able to do just that for 3.x, being major version change.

I haven't looked into PR yet (just description) to comment on changes, but wanted to point out that change as described likely can't be made in 2.x branch.

src/main/java/com/fasterxml/jackson/dataformat/xml/deser/XmlTextPropertyNameHolder.java
```diff
@@ -0,0 +1,6 @@
+package com.fasterxml.jackson.dataformat.xml.deser;
+
+public class XmlTextPropertyNameHolder {
```
cowtowncoder: I think I'd want different name for this, or (if there's natural place) just add as constant in an existing class/interface. But not sure exactly where.

src/main/java/com/fasterxml/jackson/dataformat/xml/deser/XmlTextPropertyNameHolder.java
```diff
@@ -0,0 +1,6 @@
...
+
+    public static final String PROPERTY_NAME = "<xml:text>";
```
cowtowncoder: Should probably be `DEFAULT_PROPERTY_NAME` as it can be changed.

cowtowncoder:
Ok so I do like the idea -- just need to re-create for 3.0.0 (branch `3.x`).

duoduobingbing:
> Ok so I do like the idea -- just need to re-create for 3.0.0 (branch `3.x`).

Thanks for looking into this so swiftly.

I did rebase onto `3.x` and just used `FromXmlParser.DEFAULT_TEXT_PROPERTY` and removed the newly introduced `XmlTextPropertyNameHolder.PROPERTY_NAME` completely.

src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
```diff
@@ -208,6 +209,10 @@ public PropertyName findNameForDeserialization(MapperConfig<?> config, Annotated
         PropertyName pn = PropertyName.merge(_findXmlName(a),
                 super.findNameForDeserialization(config, a));
         if (pn == null) {
+            if(_hasAnnotation(a, JacksonXmlText.class)){
+                return PropertyName.construct(FromXmlParser.DEFAULT_TEXT_PROPERTY);
```
cowtowncoder: Quick notes: not just `_hasAnnotation` -- `JacksonXmlText.value()` must be `true` (specifying `false` can be used to disable annotation's effect).

But I have related concern: there now won't be a way to override name of logical property to use as it's effectively hard-coded. Should there be a way to re-configure that? And if so, how?
(`MapperConfig` passed won't have access to XML-specific config mapper has. But maybe this introspector should have its own config or something.

src/main/java/tools/jackson/dataformat/xml/deser/FromXmlParser.java
```diff
@@ -35,10 +35,10 @@ public class FromXmlParser
     implements ElementWrappable
 {
     /**
-     * The default name placeholder for XML text segments is empty
-     * String ("").
+     * The default name placeholder for XML text segments is &lt;xml:text&gt;
+     * @since 3.0.0 Is now &lt;xml:text&gt; - was empty String ("") before.
```
cowtowncoder: nit: 3.0.0 -> 3.0 (we don't usually use patch version for API level changes)

duoduobingbing: fixed

src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
duoduobingbing: For the first note: This has been fixed in the latest commit of this pull request

For the second note: I've added a config on the introspector so that this can be configured.

src/main/java/tools/jackson/dataformat/xml/deser/FromXmlParser.java
```diff
@@ -35,10 +35,10 @@ public class FromXmlParser
     implements ElementWrappable
 {
     /**
-     * The default name placeholder for XML text segments is empty
-     * String ("").
+     * The default name placeholder for XML text segments is &lt;xml:text&gt;
+     * @since 3.0 Is now &lt;xml:text&gt; - was empty String ("") before.
      */
-    public final static String DEFAULT_UNNAMED_TEXT_PROPERTY = "";
+    public final static String DEFAULT_TEXT_PROPERTY = "<xml:text>";
```
cowtowncoder, line 45: Let's rename as "DEFAULT_XML_TEXT_PROPERTY" -- will do.

cowtowncoder: Oh. Actually, n/m -- this is at lower level, where `@XmlText` is not considered. Rather it is for "anonymous" text segments. Will slightly reword description but will not change constant name.

cowtowncoder:
Ok. I am not quite sure what the added configurability does, to be honest. I know I asked for it, but the combination of name and "infer" flag is bit unclear to me.

I realized that `FromXmlParser` already allows specifying different underlying name to use like so

```
FromXmlFactory f = FromXmlFactory.builder()
    .nameForTextElement("text-marker-element")
    .build();
```

So, f.ex input like:

```xml
<root attr="123">456</root>
```

would in 3.0 be exposed as:

```json
{
  "root": {
    "attr": "123",
    "<xml:text>": "456"
  }
}
```

with default settings.

Given this, I wonder if annotation-level configurability is really needed? If it is, how does it control things?

src/main/java/tools/jackson/dataformat/xml/deser/XmlDeserializationContext.java
```diff
@@ -68,7 +68,7 @@ public String extractScalarFromObject(JsonParser p, ValueDeserializer<?> deser,
             final String propName = p.currentName();
             JsonToken t = p.nextToken();
             if (t == JsonToken.VALUE_STRING) {
-                if (propName.equals("")) {
+                if (FromXmlParser.DEFAULT_TEXT_PROPERTY.equals(propName)) {
```
cowtowncoder: I think this needs to actually consider `FromXmlParser._cfgNameForTextElement`, since that may be configured different from default. Comment above hints at that.

Will change to do that.

src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
```diff
@@ -208,6 +215,13 @@ public PropertyName findNameForDeserialization(MapperConfig<?> config, Annotated
...
                 super.findNameForDeserialization(config, a));
         if (pn == null) {
+            JacksonXmlText jacksonXmlTextAnnotation = _findAnnotation(a, JacksonXmlText.class);
+
+            if (jacksonXmlTextAnnotation != null && jacksonXmlTextAnnotation.value() &&
+                !_cfgIntrospectorConfig.inferXmlTextPropertyName()) {
+                return _cfgIntrospectorConfig.xmlTextPropertyName();
+            }
+
             if (_hasOneOf(a, ANNOTATIONS_TO_INFER_XML_PROP)) {
```
cowtowncoder, line 225: This will look for `@JacksonXmlText` OR `@JacksonXmlElementWrapper` and handling will now differ -- former should use configured "text element name", latter should not (should return "use default").

src/test/java/tools/jackson/dataformat/xml/node/JsonNodeBasicDeserTest.java
```diff
@@ -43,7 +44,7 @@ public void testMixedContent() throws Exception
     {
         JsonNode fromXml = XML_MAPPER.readTree("<root>first<a>123</a>second<b>abc</b>last</root>");
         final ObjectNode exp = XML_MAPPER.createObjectNode();
-        exp.putArray("")
+        exp.putArray(FromXmlParser.DEFAULT_TEXT_PROPERTY)
```
cowtowncoder, line 47: Ohhhhhhh. This is NOT good -- what used to be something like:

```
{
  "":["first","second","last"],
  "a":"123",
  "b":"abc"
}
```

now looks like:

```
{
  "<xml:text>":["first","second","last"],
  "a":"123",
  "b":"abc"
}
```

which I don't think is what anyone likes to see :-(

So for `JsonNode` at least exposing XML character data sections should be with nominal key of "".
And I don't think it is reasonable to expected those using `XmlMapper.readTree()` to have explicitly configure things to work this way.

We need to figure out another way to handle the issue, I think.

src/test/java/tools/jackson/dataformat/xml/node/JsonNodeMixedContent403Test.java
```diff
@@ -19,23 +20,23 @@ public class JsonNodeMixedContent403Test extends XmlTestUtil
     public void testMixedContentBefore() throws Exception
     {
         // First, before elements:
-        assertEquals(JSON_MAPPER.readTree(a2q("{'':'before','a':'1','b':'2'}")),
+        assertEquals(JSON_MAPPER.readTree(a2q(String.format("{'%s':'before','a':'1','b':'2'}", FromXmlParser.DEFAULT_TEXT_PROPERTY))),
```
cowtowncoder, line 23: and same here obviously.

src/test/java/tools/jackson/dataformat/xml/node/JsonNodeBasicDeserTest.java
duoduobingbing: > We need to figure out another way to handle the issue, I think.

I mean this PR was just a naive shot at solving #306 by changing the property name, if it has to be `""` I'm fine with closing this PR - because I do not know how to make it so, that the property name differs from the nominal key.

src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
duoduobingbing: I do not quite understand, the highlighted line only looks for `@JacksonXmlText`

cowtowncoder: Sorry, I meant line 225 right after (`if (_hasOneOf(...)`).

src/test/java/tools/jackson/dataformat/xml/node/JsonNodeBasicDeserTest.java
cowtowncoder: Right. I'll leave this open because I'd really like to figure out a way and feel there probably is a way (despite not seeing it yet). :)

I appreciate your attempt; sorry it took me a while to sync up to what changes really mean.

duoduobingbing: No Problem, thanks for having a look at it. I will close this for now. Feel free to reopen or use the code in any way, shape or form you see fit.

cowtowncoder: Ok. Thank you for your help here @duoduobingbing .

PR #756, opened 2025-05-13
Change `FromXmlParser` to reduce mutability

cowtowncoder:


src/main/java/tools/jackson/dataformat/xml/deser/FromXmlParser.java
```diff
@@ -208,10 +210,6 @@ public Version version() {
         return PackageVersion.VERSION;
     }
 
-    public void setXMLTextElementName(String name) {
```
cowtowncoder, line 211: This is the thing remove as unnecessary, specifically... :)

Issue #735, opened 2025-03-15
Java Record with @JacksonXmlText stop working with 2.18.2


akastyka:
### Search before asking

- [x] I searched in the [issues](https://github.com/FasterXML/jackson-databind/issues) and found nothing similar.

### Describe the bug

Following class can't be deserialized with Jackson 2.18.2:
`public record Variable(
        @JacksonXmlProperty(isAttribute = true, localName = "name") String name,
        @JacksonXmlText String value) {}`

deserialization fails with:
`Invalid definition for property '' (of type com.Variable): Could not find creator property with name '' (known Creator properties: [name, value])
 at [Source: (StringReader); line: 1, column: 1]`

it works with Jackson 2.17.2. Also it works with 2.18.2 but with POJO only

### Version Information

2.18.2

### Reproduction

 -->
```java
public record Variable(
        @JacksonXmlProperty(isAttribute = true, localName = "name") String name,
        @JacksonXmlText String value) {
    public static void deserialize() throws JsonProcessingException {
        XmlMapper xmlMapper = new XmlMapper();

        Variable deserialized = xmlMapper.readValue(
                "<jackson name=\"value\">test</jackson>", Variable.class);
        System.out.println(deserialized);
    }
}
``` 

### Expected behavior

Java record is deserialized

cowtowncoder:
Closed #734 in favor of this one.

Suspecting that marker "" for `@JacksonXmlText` getting wrongly associated might be causing/related to the issue here.
Although Property introspection changes in 2.18 also likely related. But quite odd things work with POJOs with same annotations... handling differs, but 2.18 handles Records more similarly to POJOs that 2.17.

pjfanning:
> Closed [#734](https://github.com/FasterXML/jackson-dataformat-xml/issues/734) in favor of this one.
> 
> Suspecting that marker "" for `@JacksonXmlText` getting wrongly associated might be causing/related to the issue here. Although Property introspection changes in 2.18 also likely related. But quite odd things work with POJOs with same annotations... handling differs, but 2.18 handles Records more similarly to POJOs that 2.17.

https://github.com/FasterXML/jackson-dataformat-xml/issues/734#issuecomment-2729763968 seems to suggest that creating a class with the equivalent definition to a record works better in 2.18 than the record.

Aaur1s:
[...]

PR #776, opened 2025-09-14
Create XmlClassDeser735Test.java

pjfanning:
* adds test like the failing one that was added for #734 - this one works with a Java bean style class but only works if an empty constructor is added
* useful for regression testing
* the failing test still fails with latest code but now fails with this when run with Java 17
```
[ERROR]   XmlRecordDeser734Test.testDeser:25 » InvalidDefinition Invalid definition for property '' (of type `com.fasterxml.jackson.dataformat.xml.records.failing.XmlRecordDeser734Test$Amount`): Could not find creator property with name '' (known Creator properties: [value, Ccy])
 at [Source: (StringReader); line: 1, column: 1]
```

### DIFF
--- a/src/test-jdk17/java/com/fasterxml/jackson/dataformat/xml/records/failing/XmlRecordDeser735Test.java
+++ b/src/test-jdk17/java/com/fasterxml/jackson/dataformat/xml/records/failing/XmlRecordDeser735Test.java
@@ -9,8 +9,8 @@ import com.fasterxml.jackson.dataformat.xml.testutil.failure.JacksonTestFailureE
 
 import static org.junit.jupiter.api.Assertions.*;
 
-// [dataformat-xml#734]
-public class XmlRecordDeser734Test extends XmlTestUtil
+// [dataformat-xml#735]
+public class XmlRecordDeser735Test extends XmlTestUtil
 {
     record Amount(@JacksonXmlText String value,
                   @JacksonXmlProperty(isAttribute = true, localName = "Ccy") String currency) {}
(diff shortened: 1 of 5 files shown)


pjfanning:
@cowtowncoder Someone else has commented on #735 - which is related to this.

I debugged the failing record test for 734 and the immediate issue is that the PropertyDef for the 'value' field has getName=null. The getInternalName knows the right name ('value'). The new test in this PR (which has a class for Amount instead of a record) has the same issue but it seems to offer up an alternative way to set the value.

In the code snippet for BeanDeserializerFactory, the failure is thrown at line 644.
https://github.com/FasterXML/jackson-databind/blob/866e33850696ed785b1bbfc2d1dac15d188e3a1d/src/main/java/com/fasterxml/jackson/databind/deser/BeanDeserializerFactory.java#L624-L647

If the code was modified at line 630 to say that if the prop.getName returned null that we failed over to prop.getInternalName then the immediate issue would not happen because a propertyDef would be found. I haven't tested this change and something else might go wrong afterwards.

cowtowncoder:
Since #734 was closed as dup for #735 (for same issue), maybe change 734 references to 735?

pjfanning:
@cowtowncoder renamed tests to reference issue #735 

Any thoughts on changing BeanDeserializerFactory and failing over to the propDef.getInternalName when propDef.getName returns a null or empty value?

pjfanning:
@cowtowncoder the record test for 734 (renamed to 735) seems to pass if you add the param-names module.
```
        mapper.registerModule(new com.fasterxml.jackson.module.paramnames.ParameterNamesModule());
```

pjfanning:
> @cowtowncoder the record test for 734 (renamed to 735) seems to pass if you add the param-names module.
> 
> ```
>         mapper.registerModule(new com.fasterxml.jackson.module.paramnames.ParameterNamesModule());
> ```

Actually, this doesn't work. I must have made a mistake when I ran it earlier.

cowtowncoder:
I am not sure about using "getInternalName"; the name really needs to be marker (like empty String) for `@JacksonXmlText` behavior, so as a general change I don't think that would work.
If I remember correctly, internal here means "implicit" name; default before explicit (annotation) override, if any.

cowtowncoder (approved):

PR #778, opened 2025-10-08
New annotation JacksonXmlInclude works like JsonInclue but for xml.

IMurzich:
Use this annotation when need to have different result for json and xml output. Parameters for this annotation same as for JsonInclude.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
+++ b/src/main/java/com/fasterxml/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
@@ -3,6 +3,7 @@ package com.fasterxml.jackson.dataformat.xml;
 import java.lang.annotation.Annotation;
 import java.util.List;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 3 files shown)


cowtowncoder:
Thank you for your contribution.

Unfortunately I don't think we want to go down the route of per-format annotations, duplicating all functionality.

So while the idea of having a mechanism for annotating different behavior for different format is valid, we need to figure out more maintainable approach.

IMurzich:
Hi @cowtowncoder !
You absolutely true about new mechanism, but I don't see any other way for branch 2.x. Therefore I suggest stay this change in 2.x branch only and use this case (different incluthion rules for different output formats) when develop new approach.

Thanks.

pjfanning:
> Hi @cowtowncoder ! You absolutely true about new mechanism, but I don't see any other way for branch 2.x. Therefore I suggest stay this change in 2.x branch only and use this case (different incluthion rules for different output formats) when develop new approach.
> 
> Thanks.

@IMurzich 2.x is no longer the dev branch. Most changes that will go in 2.x in future will be bug fixes. This is not a bug fix.

IMurzich:
Ok, will I move this change to 3.x branch?

cowtowncoder:
At this point I wont be accepting it on any branch sorry. So can stay as is.

IMurzich:
@cowtowncoder , thank you for your time.

Issue #795, opened 2026-01-16
`HttpHeader` object (= <httpHeaders>) is not wrapped by the XML `<property>` tag


cmoulliard:
## Issue

The following Maven Settings XML
```
<settings>
    <servers>
        <server>
            <id>maven-snapshots</id>
            <configuration>
                <httpHeaders>
                    <property>
                        <name>X-JFrog-Art-Api</name>
                        <value>myApiToken</value>
                    </property>
                </httpHeaders>
            </configuration>
        </server>
    </servers>
</settings>
```
is generated as such as string during jackson serialization
```
<settings>
   <servers>
       <server>
           <id>maven-snapshots</id>
           <configuration>
               <httpHeaders>
                   <name>X-JFrog-Art-Api</name>
                   <value>myApiToken</value>
               </httpHeaders>
               <timeout>10000</timeout>
           </configuration>
       </server>
   </servers>
</settings>
```

As you can see the `<httpHeaders>` is not wrapped with `<property />`.

The code is available here: 

Here is a snapshot of the java class: https://github.com/ch007m/maven-settings-httpheader-issue/blob/main/src/test/java/dev/snowdrop/jackson/httpheader/issue/MavenSettingsSimplified.java
```
    public static class ServerConfiguration {
        @JacksonXmlProperty(
            localName = "property"
        )
        @JacksonXmlElementWrapper(
            localName = "httpHeaders",
            useWrapping = true
        )
        @JsonIgnore
        private final @Nullable List<HttpHeader> httpHeaders;
        private final @Nullable Long timeout;

        @JsonCreator
        public ServerConfiguration(@JsonProperty("httpHeaders") @Nullable List<HttpHeader> httpHeaders, @JsonProperty("timeout") @Nullable Long timeout) {
            this.httpHeaders = httpHeaders;
            this.timeout = timeout;
        }

        public @Nullable List<HttpHeader> getHttpHeaders() {
            return this.httpHeaders;
        }

        public @Nullable Long getTimeout() {
            return this.timeout;
        }

        public ServerConfiguration withHttpHeaders(final @Nullable List<HttpHeader> httpHeaders) {
            return this.httpHeaders == httpHeaders ? this : new ServerConfiguration(httpHeaders, this.timeout);
        }

        public ServerConfiguration withTimeout(final @Nullable Long timeout) {
            return this.timeout == timeout ? this : new ServerConfiguration(this.httpHeaders, timeout);
        }
    }

[...]

Issue #517, opened 2022-03-23
XML wrapper doesn't work with java records


Antolius:
Using `@JacksonXmlElementWrapper` and `@JacksonXmlProperty` annotations with Java records results in deserialization exception.

For example, I'd expect following XML:

```xml
<Request>
  <messages>
    <message>
      <text>given text</text>
    </message>
  </messages>
</Request>
```

to deserialize into following java records:

```java
public record Request(
        @JacksonXmlElementWrapper(localName = "messages")
        @JacksonXmlProperty(localName = "message")
        List<Message> messages
) {
    public Request {
    }

    private Request() {
        this(null);
    }
}

public record Message(String text) {
    public Message {
    }

    private  Message() {
        this(null);
    }
}
```

However, it results in an exception:

```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Invalid definition for property 'messages' (of type `Request`): Could not find creator property with name 'messages' (known Creator properties: [message])
 at [Source: (StringReader); line: 1, column: 10]
 ```

This is unexpected because deserialization works with equivalent java classes.

This happens with Java 17 and Jackson version 2.13.2, which is the latest one as of this writing.

I've put together a minimal project which showcases the issue: [jackson-record-showcase](https://github.com/Antolius/jackson-record-showcase).

magott:
The same seem to go for `@JacksonXmlText`

pjfanning:
Try v2.15.0-rc3. This version has better Record support.

magott:
@pjfanning did not help for `@JacksonXmlElementWrapper` nor `@JacksonXmlText`. When used on kotlin data classes at least. If it works for Records, then maybe #153 needs to be re-opened?

cowtowncoder:
@magott Kotlin Data classes are different from Records (wrt implementations); Kotlin module applies tons of magic unfortunately (some of it necessary). So if there's issue with that it may need to be opened.

One practical challenge here is tho that XML module tests cannot depend on Kotlin nor vice versa.

Refers to issue #153 is probably wrong; maybe that's for Kotlin repo issue?

magott:
@cowtowncoder yes, my bad. It was https://github.com/FasterXML/jackson-module-kotlin/issues/153 If the fix for java and kotlin is different from each other. That will need to be reopened?

cowtowncoder:
[...]

Issue #149, opened 2015-04-24
@JacksonXmlElementWrapper as a @JsonCreator parameter


dmikurube:
Do you have a plan to make @JacksonXmlElementWrapper specifiable as a parameter?

I'm wondering to use Jackson-dataformat-XML to deserialize to a immutable class instance, but @JacksonXmlElementWrapper is not available as a @JsonCreator parameter. We do a similar approach for JSON by specifying @JsonProperty on @JsonCreator.

For example:

    <body>
      <type>TYPE</type>
      <labels><label>foo</label><label>bar</label></labels>
    </body>

----

    @JacksonXmlRootElement(localName="body")
    public class Body {
        @JsonCreator
        public Body(@JsonProperty("type")
                    @JacksonXmlProperty("type")
                    String type,
                    @JsonProperty("labels")
                    @JacksonXmlElementWrapper(localName="labels")
                    @JacksonXmlProperty(localName="label")
                    Collection<String> labels) {
            // ...
        }
    
        private final String type;
        private final ArrayList<String> labels;
    }

cowtowncoder:
Usage should already work, as far as recognizing annotation goes.

However.... whether it works with respect to possible reordering needed for use with `@JsonCreator` I don't know -- there's a good chance it won't. If it does not, it is likely it will not be possible to make it work, since buffering mechanism would not work well with structural changes needed to support optional wrapping of elements.

So: if this usage does not work (which I assume is the case since you filed the issue :) ), I would suggest only passing non-collection properties via Creator, and using field or setter for collection properties.
This will work, and while not optimal, it is unlikely that passing of collection via creator for XML will be supported.

dmikurube:
Thank you for the reply. Hmm, sad news... Just adding ElementType.PARAMETER on JacksonXmlElementWrapper didn't work.

Since I don't want to make the instance mutable (non-final), I'm giving it a try to hack jackson-dataformat-xml and jackson-databind. Will send a PR when it succeeds. :)

cowtowncoder:
Right, the problem is not regarding ability to add the annotation there, or even getting it observed by Jackson. Rather, it is because handling of Creator parameters is very different from regular simple properties: since object does not yet exist, parameters typically need to be buffered. And the way XML content is supported requires a few tweaks to handling of token stream, and these are not easy to handle through buffering process.

tboettch:
[...]

PR #806, opened 2026-03-15
Fix for #615: Use ValueInstantiators to rename creator properties so they match property definitions

### LINKED ISSUE #615: Deserialization of Xml with `@JacksonXmlText` fails
This is a reproduction of #198. It was mentioned opening a new issue is preferred.

The issue is, that `@JacksonXmlText` seems to work as intended for serialization, but not for deserialization.

Hence, my reproduction of the original issue with 2.15.4:

I have the following models and tests:

```java
@JacksonXmlRootElement(localName = "ITEMROOT")
public record ItemRoot(
        @JsonProperty("Item") @JacksonXmlElementWrapper(useWrapping = false)
                List<Item> item) {
    
	public record Item(
            @JsonProperty("name") @JacksonXmlProperty(isAttribute = true) String name,
            @JacksonXmlText String value) {

        @JsonCreator
        public Item(final Map<String, String> item) {
            this(item.get("name"), item.get(""));
        }
    }
}

class Tests {

    @Test
    void testDeserializeItemRoot() throws JsonProcessingException {
        final var xmlMapper = new XmlMapper().registerModule(new ParanamerModule());
        final var itemRoot =
                new ItemRoot(
                        List.of(
                                new ItemRoot.Item("name1", "value1"),
                                new ItemRoot.Item("name2", "value2")));

        final var itemRootSerialized =
                xmlMapper.writerWithDefaultPrettyPrinter().writeValueAsString(itemRoot);

        final var itemRootXml =
                """
                <ITEMROOT>
                  <Item name="name1">value1</Item>
                  <Item name="name2">value2</Item>
                </ITEMROOT>
                """;
        assertEquals(itemRootXml, itemRootSerialized);

        final var itemRootDeserialized = xmlMapper.readValue(itemRootXml, ItemRoot.class);
        assertEquals(itemRoot, itemRootDeserialized);
    }
}
```

First, I serialize the model to verify what I actually want to deserialize is correct and then I serialize the XML again.

The tests pass because of `@JsonCreator` in `Item`. Without the annotation, I get the following error on the `xmlMapper.readValue()`:

```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Invalid definition for property '' (of type `ch.sample.model.ItemRoot$Item`): Could not find creator property with name '' (known Creator properties: [name, value])
 at [Source: (StringReader); line: 1, column: 1]
```

mcvayc:
Hi,

This PR was intended to fix #615 but in the process tests for #149, #517, #735, #795 in the `tofix` package started passing as well.  I've listed all the issues for which test are now passing in the commits, but still need to verify for those that the entire issue is fixed. 

The initial solution was co-written by Claude Code and me.  After collaboratively creating a fix, I subsequently read through the code and edited it for quality and consistency.  I believe the solution is now in decent shape, at least for #615.

Please let me know what you think.  I'm happy to make further edits.

### DIFF
--- a/src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
+++ b/src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
@@ -30,6 +30,16 @@ public class JacksonXmlAnnotationIntrospector
     };
 
     /**
+     * Marker PropertyName used for {@code @JacksonXmlText} creator parameters.
+     * Must be non-empty so {@code PotentialCreator} recognizes the parameter as named.
+     * {@link tools.jackson.dataformat.xml.deser.XmlValueInstantiators} will later rename this to the actual text element
+     * name (empty string by default).
+     *
+     * @since 3.2
+     */
+    private final static PropertyName _XML_TEXT_NAME = PropertyName.construct("#xml.text");
+
+    /**
      * For backwards compatibility with 2.0, the default behavior is
      * to assume use of List wrapper if no annotations are used.
      */
@@ -209,6 +219,14 @@ public class JacksonXmlAnnotationIntrospector
                 super.findNameForDeserialization(config, a));
         if (pn == null) {
             if (_hasOneOf(a, ANNOTATIONS_TO_INFER_XML_PROP)) {
+                // [dataformat-xml#615]: Return non-empty name so PotentialCreator
+                // treats the parameter as named; XmlValueInstantiators renames to ""
+                if (a instanceof AnnotatedParameter) {
+                    JacksonXmlText textAnn = _findAnnotation(a, JacksonXmlText.class);
+                    if (textAnn != null && textAnn.value()) {
+                        return _XML_TEXT_NAME;
+                    }
+                }
                 return PropertyName.USE_DEFAULT;
             }
         }
--- a/src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
+++ b/src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
@@ -37,7 +37,7 @@ public class JacksonXmlAnnotationIntrospector
      *
      * @since 3.2
      */
-    private final static PropertyName _XML_TEXT_NAME = PropertyName.construct("#xml.text");
+    private final static PropertyName _XML_TEXT_NAME = PropertyName.construct("&xml.text");
 
     /**
      * For backwards compatibility with 2.0, the default behavior is
--- a/src/main/java/tools/jackson/dataformat/xml/deser/XmlValueInstantiators.java
+++ b/src/main/java/tools/jackson/dataformat/xml/deser/XmlValueInstantiators.java
@@ -1,0 +1,146 @@
+package tools.jackson.dataformat.xml.deser;
+
+import java.util.*;
+
+import tools.jackson.databind.*;
+import tools.jackson.databind.deser.SettableBeanProperty;
+import tools.jackson.databind.deser.ValueInstantiator;
+import tools.jackson.databind.deser.ValueInstantiators;
+import tools.jackson.databind.introspect.AnnotatedMember;
+import tools.jackson.databind.introspect.BeanPropertyDefinition;
+import tools.jackson.dataformat.xml.annotation.JacksonXmlText;
+import tools.jackson.dataformat.xml.util.AnnotationUtil;
+
+/**
+ * {@link ValueInstantiators} implementation that renames creator properties
+ * to match the names that {@link XmlBeanDeserializerModifier#updateProperties}
+ * will give to the corresponding property definitions.
+ *<p>
+ * This is needed because {@code updateProperties()} renames property definitions
+ * (e.g., {@code @JacksonXmlText} properties to {@code ""}, wrapped collections to
+ * their wrapper name), but creator parameters retain their original names. Without
+ * this renaming, {@code BeanDeserializerFactory.addBeanProps()} cannot link property
+ * definitions to creator parameters.
+ *
+ * @since 3.2
+ */
+public class XmlValueInstantiators
+    extends ValueInstantiators.Base
+    implements java.io.Serializable
+{
+    private static final long serialVersionUID = 1L;
+
+    protected final String _cfgNameForTextValue;
+
+    public XmlValueInstantiators(String nameForTextValue)
+    {
+        _cfgNameForTextValue = nameForTextValue;
+    }
+
+    @Override
+    public ValueInstantiator modifyValueInstantiator(DeserializationConfig config,
+            BeanDescription.Supplier beanDescRef, ValueInstantiator defaultInstantiator)
+    {
+        SettableBeanProperty[] creatorProps = defaultInstantiator.getFromObjectArguments(config);
+        if (creatorProps == null || creatorProps.length == 0) {
+            return defaultInstantiator;
+        }
+
+        // Build a map of original-property-name -> new-name for renames that
+        // updateProperties() will perform on the property definitions.
+        // We need to apply the same renames to creator properties so names match.
+        Map<String, String> renames = _findPropertyRenames(config, beanDescRef);
+        if (renames.isEmpty()) {
+            // Also check for @JacksonXmlText directly on creator parameters
+            // (may not appear in bean property definitions for constructor-only classes)
+            if (!_hasXmlTextCreatorParam(creatorProps)) {
+                return defaultInstantiator;
+            }
+        }
+
+        for (int i = 0, len = creatorProps.length; i < len; ++i) {
+            SettableBeanProperty prop = creatorProps[i];
+            if (prop == null) {
+                continue;
+            }
+            final String propName = prop.getName();
+
+            // First check: direct @JacksonXmlText annotation on the creator parameter
+            JacksonXmlText textAnn = prop.getAnnotation(JacksonXmlText.class);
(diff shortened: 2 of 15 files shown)


src/main/java/tools/jackson/dataformat/xml/deser/XmlValueInstantiators.java
```diff
@@ -0,0 +1,146 @@
...
+                if (!_cfgNameForTextValue.equals(propName)) {
+                    creatorProps[i] = prop.withSimpleName(_cfgNameForTextValue);
+                }
+                continue;
+            }
+
+            // Second check: rename map from property definitions
+            String newName = renames.get(propName);
+            if (newName != null && !newName.equals(propName)) {
+                creatorProps[i] = prop.withSimpleName(newName);
```
mcvayc: Other than directly editing the array items, is there a better way to update the creator properties of the instantiator?  I didn't see a setter.

src/test/java/tools/jackson/dataformat/xml/tofix/XmlTextViaCreator306Test.java
```diff
@@ -114,15 +98,4 @@ public void testIssue306NoCtor() throws Exception
         RootWithoutConstructor rootNoCtor = MAPPER.readValue(XML, RootWithoutConstructor.class);
         assertNotNull(rootNoCtor);
     }
-
-    // [dataformat-xml#423]
-    @JacksonTestFailureExpected
-    @Test
-    public void testXmlTextViaCtor423() throws Exception
```
mcvayc, line 121: This test started passing so I split it out so it could be removed from the `tofix` package.

mcvayc: The test now resides in `src/test/java/tools/jackson/dataformat/xml/deser/XmlTextViaCtor423Test.java`

mcvayc:
I see the build errors, let me take a look.

mcvayc:
Sorry about that. I somehow missed adding a module file to the commit.

I've rebased and confirmed that everything is passing on my end.
```
[INFO] --- jacoco:0.8.14:report (report) @ jackson-dataformat-xml ---
[INFO] Loading execution data file C:\Users\chips\Documents\GitHub\mcvayc\jackson-dataformat-xml\target\jacoco.exec
[INFO] Analyzed bundle 'Jackson-dataformat-XML' with 53 classes
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  15.826 s
[INFO] Finished at: 2026-03-15T18:36:53-04:00
[INFO] ------------------------------------------------------------------------
```

src/main/java/tools/jackson/dataformat/xml/deser/XmlValueInstantiators.java
cowtowncoder: That does seem dangerous. Ideally would not tweak internal data structures nor have setters but create new instances.

I'll have to think about this one.

mcvayc:
Additionally, since the rebase it looks like test #795 now also passes so I've also moved that one out of `tofix`.

src/main/java/tools/jackson/dataformat/xml/deser/XmlValueInstantiators.java
```diff
@@ -0,0 +1,146 @@
...
+    private Map<String, String> _findPropertyRenames(DeserializationConfig config,
+            BeanDescription.Supplier beanDescRef)
+    {
+        final AnnotationIntrospector intr = config.getAnnotationIntrospector();
+        Map<String, String> renames = Collections.emptyMap();
+
+        for (BeanPropertyDefinition propDef : beanDescRef.get().findProperties()) {
+            AnnotatedMember member = propDef.getPrimaryMember();
+            if (member == null) {
+                continue;
```
cowtowncoder: Should never happen (error if it does) -- just remove the check?

src/main/java/tools/jackson/dataformat/xml/deser/XmlValueInstantiators.java
```diff
@@ -0,0 +1,146 @@
...
+            }
+            final String origName = propDef.getName();
+
+            // Check @JacksonXmlText
+            Boolean isText = AnnotationUtil.findIsTextAnnotation(config, intr, member);
+            if (isText != null && isText.booleanValue()) {
```
cowtowncoder: -> if (Boolean.TRUE.equals(isText))

src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
```diff
@@ -29,6 +29,16 @@ public class JacksonXmlAnnotationIntrospector
...
 
+    /**
+     * Marker PropertyName used for {@code @JacksonXmlText} creator parameters.
+     * Must be non-empty so {@code PotentialCreator} recognizes the parameter as named.
+     * {@link tools.jackson.dataformat.xml.deser.XmlValueInstantiators} will later rename this to the actual text element
+     * name (empty string by default).
+     *
+     * @since 3.2
+     */
+    private final static PropertyName _XML_TEXT_NAME = PropertyName.construct("#xml.text");
```
cowtowncoder: Since there is possible name conflict with actual properties, it's good that this is NOT a valid XML name. I think leading "#" guarantees that?

mcvayc: Yeah, I was worried about the same thing but Gemini assured me that: 

> No, XML cannot have a pound sign (whether referring to the British Pound symbol £ or the hash symbol #) in a tag or attribute name, as it is an illegal character according to the XML naming rules.

mcvayc: I'll see if I can find a more authoritative source

mcvayc: According to the [w3.org](https://www.w3.org/TR/xml/#NT-NameStartChar), xml names must start with one of these characters:

```
[4]   	NameStartChar	   ::=   	":" | [A-Z] | "_" | [a-z] | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x2FF] | [#x370-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D] | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF] | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]
[4a]   	NameChar	   ::=   	NameStartChar | "-" | "." | [0-9] | #xB7 | [#x0300-#x036F] | [#x203F-#x2040]
```

cowtowncoder: Having written 2 XML parsers, I concur. :)

(was thinking of using `&` but `#` is fine choice as well).

src/main/java/tools/jackson/dataformat/xml/deser/XmlValueInstantiators.java
mcvayc: I added a proposed improvement in the latest commit.  This is safer but less performant, since we copy the array even if there is no need.  Another option would be to only copy the array after we identify the first rename, but then we need to check if the array is not null each time that we rename.

src/main/java/tools/jackson/dataformat/xml/JacksonXmlAnnotationIntrospector.java
mcvayc: I update the magic value to start with `&`.

cowtowncoder (approved):
LGTM!

mcvayc:
> LGTM!

Thank you!  I found one more optimization I'd like to push.

cowtowncoder:
@mcvayc This looks very good -- thank you for contributing it! I'll add release notes and merge it for inclusion in 3.2.0!

cowtowncoder:
> > LGTM!
> 
> Thank you! I found one more optimization I'd like to push.

Ok I'll wait after updating release notes.

mcvayc:
> > > LGTM!
> > 
> > 
> > Thank you! I found one more optimization I'd like to push.
> 
> Ok I'll wait after updating release notes.

Never mind, you beat me to it with your streamlining commit. :-)

I was going to delay the array copy.

src/main/java/tools/jackson/dataformat/xml/deser/XmlValueInstantiators.java
cowtowncoder: I think this is acceptable: this does not occur for every read but only when deserializer is constructed.

cowtowncoder:
@mcvayc Ok; good to, will merge. Looks like #306 is still partially failing, but fixing 5 issues is a pretty big achievement!

cowtowncoder:
> @mcvayc Ok; good to, will merge. Looks like #306 is still partially failing, but fixing 5 issues is a pretty big achievement!

I am working on #306 -- test is buggy.

mcvayc:
> > @mcvayc Ok; good to, will merge. Looks like #306 is still partially failing, but fixing 5 issues is a pretty big achievement!
> 
> I am working on #306 -- test is buggy.

Makes sense.  

Thanks for all that you do for the Jackson projects and for being so supportive of contributions.

cowtowncoder:
> 
> Thanks for all that you do for the Jackson projects and for being so supportive of contributions.

Thank you for the kind words. It is good to get feedback like this.

And thank you for providing the fixes here!

PR #830, opened 2026-03-20
Fix #559: match @XmlValue annotated Record field with constructor parameter

### LINKED ISSUE #559: JAXB @XmlValue deserializing not working with records

cowtowncoder:
Potential fix for #559 -- not sure this makes sense.


cowtowncoder:
Not super happy with complexity of the approach... will try to simplify.

PR #878, opened 2026-07-15
Re-apply entity/DTD hardening after JDK deserialization of `XmlFactory`

Sahana2524:
**Entity/DTD hardening dropped after JDK deserialization**

`XmlFactory.readResolve()` rebuilds the `XMLInputFactory` from only its stored class name, so the reconstructed factory comes back at plain Stax defaults with external-entity and DTD processing enabled again. A mapper that was safe by default silently regains the entity-expansion/XXE surface after a serialize then deserialize roundtrip. I moved the builder's hardening into a small shared helper and call it from `readResolve` so the two factory-building paths can't drift.

Reproduces with a stock `new XmlMapper()`: reading `<!DOCTYPE foo [ <!ENTITY x "HELLO"> ]><foo>&x;</foo>` throws before serialization, but after a JDK roundtrip `&x;` expands to `HELLO`. Added a regression test alongside the existing serialization/DTD tests.

### DIFF
--- a/src/main/java/tools/jackson/dataformat/xml/XmlFactoryBuilder.java
+++ b/src/main/java/tools/jackson/dataformat/xml/XmlFactoryBuilder.java
@@ -119,6 +119,17 @@ public class XmlFactoryBuilder extends DecorableTSFBuilder<XmlFactory, XmlFactor
             // 24-Oct-2022, tatu: as per [dataformat-xml#550] need extra care
             xmlIn = XMLInputFactory.newFactory();
         }
+        return secureXmlInputFactory(xmlIn);
+    }
+
+    /**
+     * Applies the default entity/DTD hardening to a freshly created
+     * {@link XMLInputFactory}. Shared so that any code path building a factory
+     * internally (including JDK-deserialization reconstruction in
+     * {@code XmlFactory.readResolve()}) applies the same protections and the two
+     * can not drift apart.
+     */
+    protected static XMLInputFactory secureXmlInputFactory(XMLInputFactory xmlIn) {
         // as per [dataformat-xml#190], disable external entity expansion by default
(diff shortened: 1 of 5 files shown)


pjfanning:
@Sahana2524 Thanks. We require a CLA as described in https://github.com/FasterXML/jackson/blob/main/CONTRIBUTING.md#paperwork - could you provide a CLA based on this?

src/main/java/tools/jackson/dataformat/xml/XmlFactoryBuilder.java
```diff
@@ -119,6 +119,17 @@ protected static XMLInputFactory defaultXmlInputFactory(ClassLoader cl) {
...
         }
+        return secureXmlInputFactory(xmlIn);
+    }
+
+    /**
+     * Applies the default entity/DTD hardening to a freshly created
+     * {@link XMLInputFactory}. Shared so that any code path building a factory
+     * internally (including JDK-deserialization reconstruction in
+     * {@code XmlFactory.readResolve()}) applies the same protections and the two
+     * can not drift apart.
```
cowtowncoder, line 130: Add `@since 3.3`.

cowtowncoder:
Looks good overall; just need CLA (just one, good for this and #879 too) and will be good to go.

src/main/java/tools/jackson/dataformat/xml/XmlFactoryBuilder.java
Sahana2524: Done, added the tag.

Sahana2524:
Pushed the `@since 3.3` tag. On the CLA, sure. I'll fill in the standard agreement and email it to info at fasterxml dot com, one copy covering both this and #879.

cowtowncoder:
@Sahana2524 thank you! One CLA fine.

cowtowncoder (approved):

PR #879, opened 2026-07-16
Use `Locale.ROOT` for case folding in `CaseInsensitiveNameSet`

Sahana2524:
`CaseInsensitiveNameSet` folds names with the JVM default locale, so case-insensitive element matching quietly breaks for names containing `I`/`i` once the default is a locale like Turkish:
- `construct()` and `contains()` both call `String.toLowerCase()` with no locale, so a wrapper name like `ITEM` stops matching an incoming `item` when the default locale is `tr` (`"ITEM".toLowerCase()` becomes dotless `ıtem`)
- the names being matched are element/attribute names from the parsed document, and this feeds unwrapped-list wrapping via `FromXmlParser.addVirtualWrapping` when `ACCEPT_CASE_INSENSITIVE_PROPERTIES` is on, so the result ends up depending on server locale
- `XmlTokenBuffer` already matches the same names with `equalsIgnoreCase`, which is locale independent, so this helper is the odd one out

Both calls now use `Locale.ROOT`. Pure-ASCII names, the common case, are unchanged.

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -152,3 +152,7 @@ Christian Beikov (@beikov)
  * Reported #871: `XmlMapper` regression in 3.2.0: `xsi:nil` collection element on
    unwrapped collection wrongly read as null collection
   (3.3.0)
+
+@Sahana2524
+ * Fixed #879:  Use `Locale.ROOT` for case folding in `CaseInsensitiveNameSet`
+  (3.3.0)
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Looks good -- just need CLA.

Sahana2524:
Sure, I'll email the signed CLA to info@fasterxml. Saw your note on #878 that one copy covers both PRs, so that's the plan.

PR #882, opened 2026-07-21
Translate base64 name-decode errors to `StreamReadException`

Sahana2524:
When a base64 `XmlNameProcessor` is configured, reading a name that is not valid base64 lets the decoder's exception escape the parser:

    java.lang.IllegalArgumentException: Input byte[] should at least have 2 bytes for base64 bytes
        at java.base/java.util.Base64$Decoder.decode(...)
        at ...XmlNameProcessors$AlwaysOnBase64NameProcessor.decodeName(XmlNameProcessors.java:230)
        at ...deser.XmlTokenStream._decodeElementName(XmlTokenStream.java:786)
        at ...deser.FromXmlParser._nextToken(FromXmlParser.java:1421)

The base64 processors decode every incoming element/attribute name through `java.util.Base64`, which throws `IllegalArgumentException` on malformed input. That went uncaught, so callers guarding reads with `catch (JacksonException)` see a raw runtime exception instead.

The base64 processors in `XmlNameProcessors` now translate the decoder's `IllegalArgumentException` into a `StreamReadException` at the point of decode, so the failure is handled where it happens and cannot leak through any parser path (root-name decode in `XmlTokenStream.initialize()` or later names in `_nextToken()`). Valid names and the default passthrough processor are unchanged. Added regression tests in `XmlNameEscapeTest` for both the prefixed and always-on base64 processors.

### DIFF
--- a/src/main/java/tools/jackson/dataformat/xml/XmlNameProcessors.java
+++ b/src/main/java/tools/jackson/dataformat/xml/XmlNameProcessors.java
@@ -3,8 +3,6 @@ package tools.jackson.dataformat.xml;
 import java.util.Base64;
 import java.util.regex.Pattern;
 
-import tools.jackson.core.exc.StreamReadException;
-
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
@@ -209,7 +207,7 @@ public final class XmlNameProcessors
             if (name.localPart.startsWith(_prefix)) {
                 String localName = name.localPart;
                 localName = localName.substring(_prefix.length());
-                name.localPart = _decodeBase64Name(BASE64_DECODER, localName);
+                name.localPart = new String(BASE64_DECODER.decode(localName), UTF_8);
             }
         }
     }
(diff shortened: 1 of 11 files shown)


cowtowncoder:
Test makes sense and problem should be addressed. But not in "nextToken()" calls -- needs to be closer to the actual fail, NameProcessor must translate failures and not leak them through.

Sahana2524:
Makes sense. Moved the translation into the base64 processors themselves: `decodeName` now wraps the decoder's `IllegalArgumentException` in a `StreamReadException`, and the parser-side catches are reverted. That covers both the root-name path and `_nextToken()` from the same place. Tests unchanged, full suite still passes.

src/main/java/tools/jackson/dataformat/xml/XmlNameProcessors.java
```diff
@@ -227,7 +229,20 @@ public void encodeName(XmlName name) {
...
         @Override
         public void decodeName(XmlName name) {
-            name.localPart = new String(BASE64_DECODER.decode(name.localPart), UTF_8);
+            name.localPart = _decodeBase64Name(BASE64_DECODER, name.localPart);
+        }
+    }
+
+    // Base64 decoder throws IllegalArgumentException on malformed input; must
+    // translate into Jackson exception so it does not leak through parser API
+    private static String _decodeBase64Name(Base64.Decoder decoder, String encoded) {
```
cowtowncoder: Oh I see -- we do not pass enough context to give access to Location info. Need to see where to change it...

cowtowncoder (approved):

Issue #350, opened 2019-07-18
Wrap Xerces/Stax (JDK-bundled) exceptions during parser initialization


cowtowncoder:
Although XML module works best when used over Woodstox or Aalto, ideally any Stax parser implementation should work for basic usage. So use of JDK-bundled Stax processor (which is nowadays based on Xerces it seems) should be supported.

But although all problems should be reported as `XMLStreamException`s, there are reports that some issues with DOCTYPE declaration are instead reported using `java.util. MissingResourceException` (subtype of `RuntimeException`).
This can lead to leakage from method `XmlFactory._initializeXmlReader()`.

It's not unclear what'd be the best approach balancing minimal intervention and safety, but for now maybe wrap all `RuntimeException`s from that particular spot, to reduce the likelihood that caller leaks exception.

PR #883, opened 2026-07-22
Translate `XMLStreamReader` creation errors to `StreamReadException`

Sahana2524:
XmlFactory._createParser catches only XMLStreamException around createXMLStreamReader:
- non-Woodstox Stax impls (SJSXP, #618) parse the XML declaration during reader creation and can throw ArrayIndexOutOfBoundsException on malformed input
- that unchecked exception escapes readValue/readTree instead of a StreamReadException, so callers catching JacksonException never see it
- the sibling paths already translate this (_initializeXmlReader for #350, the reader.next() adapter for #618); the create boundary was the gap

Wrapped it in all four _createParser overloads; the test uses a stand-in factory that throws at reader creation.

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -166,3 +166,5 @@ Christian Beikov (@beikov)
   (3.3.0)
  * Fixed #882: Translate base64 name-decode errors to `StreamReadException`
   (3.3.0)
+ * Fixed #883: Translate `XMLStreamReader` creation errors to `StreamReadException`
+  (3.3.0)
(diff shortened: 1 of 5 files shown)


cowtowncoder:
TBH, this sounds like SJSXP bug we should not pave over -- it really should not leak AAOOBEs.

I'll need to think about this a bit...

Sahana2524:
Agreed, the root cause is SJSXP and ideally it gets fixed there. The reason I went this route is that the next() path already translates the exact same AIOOBE from the SJSXP decoder (Stax2JacksonReaderAdapter, from #618), so createXMLStreamReader() looked like the one remaining boundary where it still escapes as a raw runtime exception. No urgency on my end, happy to adjust if you'd rather handle it differently.

cowtowncoder (approved):

cowtowncoder:
@Sahana2524 actually I agree with your assessment after thinking it through: this is one-off on construction and there's earlier precedents. Thank you!

PR #890, opened 2026-08-08
Update token on 19-digit `long` coercion in `isExpectedNumberIntToken`

Sahana2524:
`isExpectedNumberIntToken()` skips the token-count update for one integer-length branch:
- the `len == 19` in-`long`-range case sets `_currToken` directly, while the <=9, <=18 and BigInteger branches (and 2.x) all go through `_updateToken()`
- so coercing a 19-digit `long` to `VALUE_NUMBER_INT` is not counted, leaving `currentTokenCount()` and the `maxTokenCount` read constraint short for those values

Routed it through `_updateToken()` like the sibling branches; no effect when token counting is off.


src/test/java/tools/jackson/dataformat/xml/dos/TokenCountNumberCoercionTest.java
```diff
@@ -0,0 +1,61 @@
+package tools.jackson.dataformat.xml.dos;
```
cowtowncoder: Not related to DoS aspects, will find diff package.

cowtowncoder: Actually, one can argue it is a good package, will see if it can be added to an existing test class instead.

cowtowncoder:
Wait -- WHY would we want to double-count token count for coercion? This seems like a bug; there is no new token for coercing already-decoded token; only its token type changes.

So I think this is wrong way around; double-counting should be removed instead.

cowtowncoder (changes requested):
Wrong change -- double-counting should be eliminated, if anything; token count should be based on actual tokens, not including virtual coercion which simply changes type of already read token.

cowtowncoder (approved):
Ok, I changed my mind actually: although I think double-counting is wrong, change here is to unify handling to be consistent -- and trying to undo double-counting is a much bigger effort. So let's do that after all.

Sahana2524:
Thanks for merging. Agreed the coercion double-counting itself is questionable; consistency across the branches seemed like the safer first step, and happy to help if you want to revisit the counting semantics separately.

cowtowncoder:
@Sahana2524 right. I think double-counting is ok for now, we can tackle that in future if it becomes an issue.

PR #891, opened 2026-08-10
Enforce `StreamReadConstraints.maxNestingDepth` in `FromXmlParser`

Sahana2524:
FromXmlParser increments read-context nesting depth on each child Object/Array, but never validates it, so a configured StreamReadConstraints.maxNestingDepth is ignored when reading XML.
- ToXmlGenerator already validates StreamWriteConstraints.maxNestingDepth on the write side, and jackson-core JsonParserBase does the same on read, so the XML read path was the outlier
- the default limit (500) sits below Woodstox element-depth default (1000), so three existing deep-nesting tests reached the jackson-core limit first; I raised their maxNestingDepth so each still exercises the limit it was written for (Woodstox element depth, token count)

New test covers rejection past a low configured limit while staying within the Stax element-depth limit.

### DIFF
--- a/src/test/java/tools/jackson/dataformat/xml/dos/DeepNestingParserTest.java
+++ b/src/test/java/tools/jackson/dataformat/xml/dos/DeepNestingParserTest.java
@@ -16,12 +16,14 @@ public class DeepNestingParserTest extends XmlTestUtil
 {
     // Default StreamReadConstraints.maxNestingDepth (500) is now enforced by
     // FromXmlParser itself, before the underlying Stax implementation's own
-    // element-depth limit would kick in
+    // element-depth limit would kick in. Depth chosen just past 500 (rather
+    // than e.g. 1050) so the test pins the actual enforced boundary instead
+    // of merely confirming failure somewhere past both limits.
     @Test
     public void testDeepDoc() throws Exception
     {
         final XmlMapper xmlMapper = newMapper();
-        final String XML = createDeepNestedDoc(1050);
+        final String XML = createDeepNestedDoc(510);
         try (JsonParser p = xmlMapper.createParser(XML)) {
             while (p.nextToken() != null) { }
             fail("expected StreamConstraintsException");
--- a/src/test/java/tools/jackson/dataformat/xml/dos/DeepNestingParserTest.java
+++ b/src/test/java/tools/jackson/dataformat/xml/dos/DeepNestingParserTest.java
@@ -3,18 +3,20 @@ package tools.jackson.dataformat.xml.dos;
 import org.junit.jupiter.api.Test;
 
 import tools.jackson.core.JsonParser;
-import tools.jackson.core.exc.StreamReadException;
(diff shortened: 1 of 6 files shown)


src/test/java/tools/jackson/dataformat/xml/dos/DeepNestingParserTest.java
```diff
@@ -3,22 +3,28 @@
...
 {
+    // The underlying Stax impl (Woodstox) enforces its own element-depth limit;
+    // raise the jackson-core nesting limit here so this test keeps exercising
+    // that Woodstox limit rather than the (now enforced) jackson-core one.
     @Test
     public void testDeepDoc() throws Exception
     {
-        final XmlMapper xmlMapper = newMapper();
+        final XmlMapper xmlMapper = mapperBuilder(XmlFactory.builder()
+                .streamReadConstraints(StreamReadConstraints.builder()
```
cowtowncoder: I don't understand this -- why are we testing Woodstox enforced limits at all? Tests here should be about XML module, not underlying Stax implementation.

cowtowncoder:
Makes sense if update was indeed missing. But if so, we'd need this against 3.1 branch.

src/test/java/tools/jackson/dataformat/xml/dos/DeepNestingParserTest.java
Sahana2524: Fair point. The old test was effectively pinned to Woodstox's element-depth limit and I kept it that way by raising the jackson-core limit. Changed it to use default constraints instead, so it now verifies the module's own nesting-depth check (which fires at 500, before Woodstox's 1000 would). The only remaining raised limit in dos/ is TokenCountTest, where it just keeps the token-count check as the one that triggers for the deep doc.

Sahana2524:
Rebased onto 3.1 and retargeted the PR. Only difference from the 3.x version is one fewer call site, since the root-wrap code isn't on 3.1. And yes, the check was missing: with the parser change reverted, the new low-limit test parses a 50-deep doc to completion despite maxNestingDepth(10) being configured. Full suite is green on 3.1.

cowtowncoder:
Thank you @Sahana2524 !

cowtowncoder:
I am still bit surprised about the gap -- I thought nesting limit on parsing was even tested.

But what I _think_ happened is that there is

    tools.jackson.dataformat.xml.dos.DeepNestingParserTest

which relied on _Woodstox_ limit, not XML module. But that should not be done for multiple reasons (f.ex diff parsers; configuring stricter Jackson limit etc).

cowtowncoder (approved):
LGTM: will validate later tonight for sure before merging.

cowtowncoder (approved):

cowtowncoder:
Fixed in 3.1, backported to 2.x so will be included in 

* 2.18.10, 2.21.6, 2.22.2
* 3.1.6, 3.2.2

PR #893, opened 2026-08-11
Decode element name before matching virtual wrapper in `_initStartElement`

Sahana2524:
Unwrapped repeated elements are matched against the active virtual wrapper by raw element name, but the wrapper holds the name after `XmlNameProcessor` decoding (`repeatStartElement` stores the decoded `_localName`/`_namespaceURI`).

- with a base64 (or always-on base64) name processor, a repeated `<encoded>` element never matches its own wrapper, so each item closes the virtual array and only the last one survives
- reading a 3-item `useWrapping=false` list under the always-on base64 processor silently yields a 1-item list, with no exception to signal the loss
- `_handleRepeatElement` and `_handleEndElement` already operate on decoded names, so `_initStartElement` was the only spot comparing before decoding

Fix decodes the name up front and matches on the decoded value; the delayed-replay branch keeps the raw name since `_handleRepeatElement` decodes it again downstream. Added round-trip tests for the always-on and colliding-prefix base64 processors (both fail on the current tree).

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -177,3 +177,6 @@ Christian Beikov (@beikov)
   (3.3.0)
  * Fixed #890: Count 19-digit `long` coercion in `isExpectedNumberIntToken()`
   (3.3.0)
+ * Fixed #893: Decode element name before matching virtual wrapper in
+   `_initStartElement`
+  (3.3.0)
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Makes sense. Would this make sense to target 3.1 or 3.2 (I forget when element name encoding feature was added)?

Sahana2524:
The name processor feature goes back to 2.14 (#531), so it predates 3.0 entirely -- I checked and the 2.x branch has the same raw-name `matchesWrapper()` comparison in `_initStartElement()`. So 3.1 is a valid target, and in principle this could be backported as far as 2.14. Happy to rebase onto whichever branch you'd like it to land in.

cowtowncoder (approved):

PR #899, opened 2026-08-18
Return `null` from `nextStringValue()` at end of input

Sahana2524:
java.lang.IllegalStateException: Internal error: unrecognized XmlTokenStream token: 8

nextStringValue() runs the XmlTokenStream event through a switch, but the XML_END branch has no break and falls into the default: case that throws. So one call past the last token raises an unchecked exception instead of reporting end-of-input as null, the way nextToken() already does for XML_END. Callers that only catch JacksonException do not see it.

Added the missing break so the XML_END case reaches the method's shared `return null`. Extended the existing `XmlParserNextXxxTest` with one call past the end, which throws on the current tree and now returns null.

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -183,3 +183,6 @@ Christian Beikov (@beikov)
  * Fixed #893: Decode element name before matching virtual wrapper in
    `_initStartElement`
   (3.3.0)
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Will also backport in 2.x (for 2.3.0).

PR #901, opened 2026-08-20
Include `ENTITY_REFERENCE` replacement text in element values (instead of silently dropping it)

Sahana2524:
Running the new `EntityReferenceReadTest` against the unpatched tree:

    EntityReferenceReadTest.testDeclaredEntityNotReplacedByReader:56 expected: <fooxxbar> but was: <foobar>
    EntityReferenceReadTest.testUndeclaredEntityNotReplacedByReader:73 Expected tools.jackson.core.exc.StreamReadException to be thrown, but nothing was thrown.

With a reader that has `IS_REPLACING_ENTITY_REFERENCES` disabled (Woodstox honors it, and the same goes for any reader handed to `createParser(XMLStreamReader)`), `&e;` inside element text arrives as an `ENTITY_REFERENCE` event, and `XmlTokenStream` only collects `CHARACTERS`/`CDATA`, so the reference just vanished from the value with no error. Now it contributes the reader's replacement text, and fails with a located `StreamReadException` when the reader has none (undeclared entity), which seemed better than a quietly truncated string. Default readers expand entities themselves and never produce these events, so nothing changes there. The same code is on 2.x and 3.1; happy to retarget if you would rather have it land there first.

### DIFF
--- a/src/main/java/tools/jackson/dataformat/xml/deser/XmlTokenStream.java
+++ b/src/main/java/tools/jackson/dataformat/xml/deser/XmlTokenStream.java
@@ -644,6 +644,9 @@ public class XmlTokenStream
                 // note: SPACE is ignorable (and seldom seen), not to be included
                 case XMLStreamConstants.CHARACTERS:
                 case XMLStreamConstants.CDATA:
+                // Only reported if the reader does not replace entity references
+                // (or could not expand this one): still part of text content
+                case XMLStreamConstants.ENTITY_REFERENCE:
                     // 17-Jul-2017, tatu: as per [dataformat-xml#236], need to try to...
                     {
                         String str = _getText(_xmlReader);
@@ -684,6 +687,7 @@ public class XmlTokenStream
             // note: SPACE is ignorable (and seldom seen), not to be included
             case XMLStreamConstants.CHARACTERS:
             case XMLStreamConstants.CDATA:
+            case XMLStreamConstants.ENTITY_REFERENCE:
                 {
                     String str = _getText(_xmlReader);
                     if (chars == null) {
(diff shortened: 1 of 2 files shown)


pjfanning:
entity references are really frowned on - turned out be a real security issue - so adding back support for them is probably not something that we would likely want to backport

anything that gets added needs to be thoroughly tested to avoid reopening security attack vectors

cowtowncoder:
@pjfanning This does not enable entity reference resolution, just handles them better if user explicitly enables. So should not be problematic wrt security... although review by coding agent can help evaluate that angle just to be sure.

I think throwing exception when encountering unresolved entity ref might be net positive for security?

Sahana2524:
Fair point on the backport, keeping this on 3.x only is fine by me.

On the security angle, I went through the attack-vector cases against this branch with Woodstox 7.2.2. The change never resolves anything itself, it only consumes whatever the reader already hands back from `getText()`, and `ENTITY_REFERENCE` events only exist once someone has explicitly turned `IS_REPLACING_ENTITY_REFERENCES` off (or wrapped their own reader). With the module's default factory nothing changes: the replacing reader rejects `&e;` with "Undeclared general entity" before this code is reached, DTD or not. With a non-replacing reader an external entity (`<!ENTITY ext SYSTEM "file:///...">`) comes through with a null replacement text, so it's now a `StreamReadException` instead of quietly vanishing; the file content never shows up, and that held even with `IS_SUPPORTING_EXTERNAL_ENTITIES` enabled since Woodstox doesn't expand externals in non-replacing mode. Nested internal entities (`<!ENTITY b '&a;&a;'>`) come back as the literal one-level replacement text, so there's no recursive expansion or amplification in this path; the replacing path, with Woodstox's own entity limits, is untouched.

I've pushed two tests pinning the external-entity and nested-entity cases so that stays true. Agree that failing on an unresolved reference is the better outcome than the old silently truncated string.

cowtowncoder:
Note: ok for 3.x, no backporting to maintenance branches.

cowtowncoder (approved):

cowtowncoder (approved):

PR #903, opened 2026-08-24
Split embedded `]]>` across CDATA sections when serializing

Sahana2524:
**CDATA values containing `]]>` fail to serialize**

A `@JacksonXmlCData` string is handed straight to `writeCData`, and since `]]>` closes a CDATA section the underlying Woodstox writer rejects it, so any value carrying that sequence cannot be written at all and the round-trip guarantee breaks. The safe form is to split at each `]]>` (keep the `]]` in one section, push the `>` into the next); a coalescing reader, which is the default here, reads the pieces back as the original text. I put the split in a shared `StaxUtil.writeCData` helper so the plain and pretty-printing paths behave the same.

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -189,3 +189,5 @@ Christian Beikov (@beikov)
  * Fixed #899: Return `null` from `nextStringValue()` at end-of-input (instead
    of throwing `IllegalStateException`)
   (3.3.0)
+ * Fixed #903: Split embedded `]]>` across CDATA sections when serializing
+  (3.3.0)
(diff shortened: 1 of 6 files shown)


Sahana2524:
any update?

cowtowncoder:
I am rather overloaded -- will eventually get to this one, not forgotten.

cowtowncoder:
Too bad we cannot rely on Woodstox setting

    factory.setProperty(WstxOutputProperties.P_OUTPUT_FIX_CONTENT, Boolean.TRUE);

which would do the same. But that'd only work with Woodstox, not other Stax impls.

So let's go with this instead.

cowtowncoder (approved):

PR #905, opened 2026-08-27
Guard `ToXmlGenerator` cast in `serializeWithType()`/`_serializeObjectId()` for `convertValue()`

Sahana2524:
`ClassCastException` from `XmlMapper.valueToTree()` / `convertValue()` on any `@JsonTypeInfo(include = As.PROPERTY)` bean (the default inclusion):

    java.lang.ClassCastException: class tools.jackson.databind.node.TreeBuildingGenerator
        cannot be cast to class tools.jackson.dataformat.xml.ser.ToXmlGenerator

`serializeWithType()` and `_serializeObjectId()` cast the generator to `ToXmlGenerator` to flag the type id as an attribute, but those two conversion entry points serialize into a `TokenBuffer` / tree generator rather than a `ToXmlGenerator`. The sibling methods `_serializeProperties()` and `_serializePropertiesFiltered()` in the same class already skip the XML-specific handling with an `instanceof ToXmlGenerator` guard (comment: "During 'convertValue()', need to skip"); these two type-id paths were the only ones missing it, so I added the same guard to both.

Real XML serialization is unchanged: the type id is still written as an attribute. Added a small regression test covering `valueToTree`, `convertValue`, and the unchanged XML output.

### DIFF
--- a/.github/workflows/coverage-comment.yml
+++ b/.github/workflows/coverage-comment.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
     - name: Download comment artifact
       id: download
-      uses: dawidd6/action-download-artifact@b6e2e70617bc3265edd6dab6c906732b2f1ae151 # v21
+      uses: dawidd6/action-download-artifact@d63b86af1b34672e53c440b1b83979861906bad7 # v24
       with:
         name: pr-comment
         path: pr-comment/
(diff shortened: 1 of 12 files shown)


Sahana2524:
any update?

cowtowncoder (approved):

PR #907, opened 2026-08-29
Recompute XML metadata for filtered props in `XmlBeanSerializerBase`

Sahana2524:
**Wrong-property XML metadata after `@JsonIgnoreProperties`**

`XmlBeanSerializerBase` keeps four index-parallel structures (`_attributeCount`, `_textPropertyIndex`, `_xmlNames`, `_cdata`) that the base constructor derives from `_props`. The `(src, toIgnore, toInclude)` and `(src, properties, filteredProperties)` constructors copied them straight from `src`, but `super(...)` there rebuilds `_props` with a shorter or re-ordered set, so every slot after a dropped or moved property lands on the wrong one. A member-level `@JsonIgnoreProperties`/`@JsonIncludeProperties` (which routes through `withByNameInclusion`) or a modifier that hands back `withProperties(...)` is enough to hit it, and nothing throws: a plain element gets emitted as an attribute, a `@JacksonXmlText` value swaps places with the next element, or CDATA attaches to a neighbor, so the output no longer reads back to the same object.

Moved the derivation into a shared helper and recomputed it against the rebuilt array in those two constructors; the copy-from-`src` path stays for the constructors that keep the same property set. Added regression tests for the attribute, text, and CDATA cases (each fails before the change). Full suite green under `./mvnw -B -ff -ntp verify`.

### DIFF
--- a/src/main/java/tools/jackson/dataformat/xml/ser/XmlBeanSerializerBase.java
+++ b/src/main/java/tools/jackson/dataformat/xml/ser/XmlBeanSerializerBase.java
@@ -62,47 +62,11 @@ public abstract class XmlBeanSerializerBase extends BeanSerializerBase
     public XmlBeanSerializerBase(BeanSerializerBase src)
     {
         super(src);
-
-        // Then make sure attributes are sorted before elements, keep track
-        // of how many there are altogether
-        int attrCount = 0;
-        for (BeanPropertyWriter bpw : _props) {
-            if (_isAttribute(bpw)) { // Yup: let's build re-ordered list then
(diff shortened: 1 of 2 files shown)


Sahana2524:
gentle ping

pjfanning:
@Sahana2524 there are a lot of PRs to review. Your ones look good to me but it could take another few days or weeks for them to be merged. If you need the fixes urgently, you'll need to build your own jar for the time being or embed the code in your own apps.

cowtowncoder (approved):

PR #909, opened 2026-09-01
Clear forced `xsi:type` attribute state in `ToXmlGenerator.writeName`

Sahana2524:
`writeName`'s synthetic `xsi:type` handling forces attribute mode and the XSI namespace onto the next value, and only the bean-serializer path clears it (by assigning the next name through `setNextName`):

- a `Map` / `JsonNode` / `@JsonAnyGetter` key of `xsi:type` leaves every following sibling as an `xsi:`-prefixed attribute instead of an element
- `{"xsi:type":"A","type":"B"}` emits a duplicate `xsi:type` attribute, so the module produces XML it can no longer read back
- `{"xsi:type":"T","nil":"true"}` emits `xsi:nil="true"`, so the whole document reads back as `null`

Clears the forced state on the next `writeName` for the content-driven case; the bean path already reset it via `setNextName`, so legitimate polymorphic output is unchanged.

### DIFF
--- a/src/main/java/tools/jackson/dataformat/xml/ser/ToXmlGenerator.java
+++ b/src/main/java/tools/jackson/dataformat/xml/ser/ToXmlGenerator.java
@@ -192,6 +192,17 @@ public class ToXmlGenerator
     protected boolean _nextIsCData = false;
 
     /**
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Thank you, @Sahana2524 !

PR #910, opened 2026-09-08
verify Stax factory type before instantiating in XmlFactory.readResolve()

Sahana2524:
**Unverified class instantiation in `XmlFactory.readResolve()`**

The Stax factory class names come back off the JDK-serialized stream (`readObject` reads them via `readUTF`), but both were handed straight to `Class.forName(name).getDeclaredConstructor().newInstance()`, so a tampered `XmlFactory`/`XmlMapper` runs the static initializer and no-arg constructor of any class named there before the cast to `XMLInputFactory`/`XMLOutputFactory` gets a chance to reject it. This is the same JDK-deserialization boundary #878 hardened for entity/DTD settings, just a different failure on it. Now each name is loaded without initializing (`Class.forName(name, false, loader)`), checked with `asSubclass(...)`, and only then constructed, so an unexpected type is refused before it can run; real factory class names round-trip exactly as before.

`2.x` has the same two call sites in its `readResolve()`, so happy to rebase this onto whichever branch you'd prefer it to land in.

AI tooling was used to help prepare this change.

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -193,3 +193,6 @@ Christian Beikov (@beikov)
  * Fixed #899: Return `null` from `nextStringValue()` at end-of-input (instead
    of throwing `IllegalStateException`)
   (3.3.0)
+ * Fixed #910: Verify Stax factory type before instantiating in
+   `XmlFactory.readResolve()`
+  (3.3.0)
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Replaced by #911.

PR #913, opened 2026-09-18
Keep `Builder.defaultUseWrapper()` from modifying shared introspector

Sahana2524:
**`defaultUseWrapper()` on a rebuilt builder also reconfigures the original mapper**

The builder applied the setting through `setDefaultUseWrapper()` on an introspector instance it shares with the mapper it came from via `rebuild()` (or one it already built), so `mapper.rebuild().defaultUseWrapper(false).build()` flips wrapping on the original too, `XmlMapper.shared()` included: types it has not cached yet are written unwrapped, and a wrapped document reads back as an empty list without any error. It now swaps in a re-configured copy, the way `nameForTextElement()` already does for the factory; the copy uses `clone()` so a custom sub-class keeps its type, and a paired introspector is re-paired from `allIntrospectors()` in the same precedence order, which seemed the smallest option without new databind API but I'm glad to reshape it.


cowtowncoder (approved):

cowtowncoder:
Changed a bit, will merge: thank you again @Sahana2524 !

PR #914, opened 2026-09-20
Apply format features set via `XmlFactoryBuilder` to `XmlFactory`

Sahana2524:
**`XmlFactory.builder()` silently drops `XmlReadFeature`/`XmlWriteFeature` changes**

The `enable()`/`disable()`/`configure()` overloads here update two builder-local masks left over from the 2.x builder that nothing reads, while `XmlFactory(XmlFactoryBuilder)` takes its masks from the base `TSFBuilder` fields, so `XmlFactory.builder().disable(XmlReadFeature.AUTO_DETECT_XSI_TYPE).build()` (or `PROCESS_XSI_NIL`, `WRITE_XML_DECLARATION`, ...) still yields a factory, and any mapper built on it, running with the defaults and no hint that the setting was lost. They now write to the base-class masks like `XmlMapper.Builder` does; I dropped the two unread `protected` fields instead of keeping them in sync, which seemed cleaner, but I can leave them in as deprecated if you would rather not remove them on a patch branch (`3.1` has the same code, `2.x` is not affected).


cowtowncoder:
Removing dup fields much better, thanks!

